PR #4, opened 2018-01-18
REPL mode

adamnfish:
cc @aug24

This PR adds an interactive REPL to the program, initiated with the `--interactive` / `-I` flag.

Instead of running a single command and printing it to the console, it will create an interactive terminal window that shows the instances being targeted and allows the user to input commands. You can view the output for each instance.

This is a working first draft. It's pretty handy for a single instance but the layout needs to be improved (uses linear layout, grid layout would be better). That would also allow us to detect the terminal size and show multiple instances side-by-side (tiled) where possible. It would also be super helpful to have command history available which leads to questions like "can we have readline support". I think this is already useful as-is though, so we can see if it gets used before making it perfect.

Picture > 1K words.

annotated-screenshot


aug24:
+1

PR #6, opened 2018-01-22
Add ssh temp key generation, tainting, tests

aug24:


src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+      case e:NoSuchProviderException => Left(FailedAttempt(
+        Failure(s"Unable to create key pair with that provider", "Error creating key", UnhandledError, None, Some(e))
+      ))
+    }
+  }
+
+  def addTaintedCommand(name: String): String = {
+    s"""
+       | [[ -f /etc/update-motd.d/99-tainted ]] || /bin/echo -e '#!/bin/bash' | /usr/bin/sudo /usr/bin/tee -a /etc/update-motd.d/99-tainted >> /dev/null;
+       | /bin/echo -e 'echo -e "\033[0;31mThis instance was tainted by ${name} at ${Calendar.getInstance().getTime()}\033[0;30m"' | /usr/bin/sudo /usr/bin/tee -a /etc/update-motd.d/99-tainted >> /dev/null;
```
adamnfish: Is "tainted" a term that is widely understood? That is what we're describing here so the file name is fine, but would this be better worded more accessibly for the user-facing message? (I don't know)

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def sshCmd(tempFile: File, instance: Instance)(implicit ec: ExecutionContext): (InstanceId, Either[CommandStatus, CommandResult]) = {
+    val cmd = s"""
+      | # Execute the following command within the next $delay seconds:
+      | ssh -i ${tempFile.getCanonicalFile.toString} ubuntu@${instance.publicIpAddressOpt.get};
+      |""".stripMargin
+    instance.id -> Right(CommandResult(cmd, ""))
+  }
+
+  def sshCmds(tempFile: File, instances: List[Instance])(implicit ec: ExecutionContext): List[(InstanceId, Either[CommandStatus, CommandResult])] = {
```
adamnfish: `sshCmd` and `sshCmds` do very different things, but are named as if one is the multi-instance version of the other. It sounds as if the implementation of `sshCmds` should be `instances.map(sshCmd)`. Could these be named differently?

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+      | /bin/sleep $delay;
+      | /bin/sed -i '/${authKey.replaceAll("/", "\\\\/")}/d' /home/ubuntu/.ssh/authorized_keys;
+      |""".stripMargin
+
+  def sshCmd(tempFile: File, instance: Instance)(implicit ec: ExecutionContext): (InstanceId, Either[CommandStatus, CommandResult]) = {
+    val cmd = s"""
+      | # Execute the following command within the next $delay seconds:
+      | ssh -i ${tempFile.getCanonicalFile.toString} ubuntu@${instance.publicIpAddressOpt.get};
+      |""".stripMargin
+    instance.id -> Right(CommandResult(cmd, ""))
```
adamnfish: Is this function mixing up the UI with the logic? It's a minor point and you may disagree, but I wonder if this file should be producing the relevant information (key file and instance IP) and the UI should be deciding how that gets displayed to the user.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  def sshCmds(tempFile: File, instances: List[Instance])(implicit ec: ExecutionContext): List[(InstanceId, Either[CommandStatus, CommandResult])] = {
+    instances.map(i => sshCmd(tempFile, i))
+  }
+
+  def getSingleInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] =
```
adamnfish: Is this really to do with `SSH` functionality? This function feels like it might be in the wrong place.

src/main/scala/com/gu/ssm/aws/EC2.scala
```diff
@@ -31,10 +31,24 @@ object EC2 {
     handleAWSErrs(awsToScala(client.describeInstancesAsync)(request).map(extractInstances))
   }
 
+  def resolveInstanceIds(ids: List[InstanceId], client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[List[Instance]] = {
```
adamnfish: I mentioned above that while this function looks reasonable on the face of it, the way it's used is probably quite surprising to the user.

src/main/scala/com/gu/ssm/aws/EC2.scala
```diff
@@ -31,10 +31,24 @@ object EC2 {
...
+    handleAWSErrs(awsToScala(client.describeInstancesAsync)(request).map(extractInstances))
+  }
+
   private def extractInstances(describeInstancesResult: DescribeInstancesResult): List[Instance] = {
     (for {
       reservation <- describeInstancesResult.getReservations.asScala
       awsInstance <- reservation.getInstances.asScala
-    } yield Instance(awsInstance.getInstanceId)).toList
+      instanceId = awsInstance.getInstanceId
+      ipAddress = awsInstance.getPublicIpAddress match {
```
adamnfish: If this is to catch the `null` case then you can do that more simply with `Option`. The `Option` constructor does a `null` check and turns `null` into `None`. So this would be the same as:

    Option(awsInstance.getPublicIpAddress)

This behaviour exists on Option precisely to make this sort of Java-land interoperability as easy as possible.

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
```diff
@@ -0,0 +1,81 @@
...
+      dos.write (rsaPublicKey.getModulus.toByteArray)
+      val publicKeyEncoded = new String (Base64.encodeBase64 (byteOs.toByteArray) )
+      "ssh-rsa " + publicKeyEncoded + " " + description
+  }
+}
+
+  import org.apache.commons.codec.binary.Base64
+  import java.io.ByteArrayOutputStream
+
+//  val rsaPublicKey = key.asInstanceOf[RSAPublicKey]
```
adamnfish: O_O

(assume this isn't meant to be here?)

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
```diff
@@ -0,0 +1,81 @@
...
+  def write(): Unit = {
+    val pemWriter = new PemWriter(new OutputStreamWriter(new FileOutputStream(file)))
+    try
+      pemWriter.writeObject(this.pemObject)
+    finally pemWriter.close()
+    println("Wrote key to file (" + file.getCanonicalFile.toString) + ")"
+  }
+}
+
+import org.apache.commons.codec.binary.Base64
```
adamnfish: imports to the top!

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
```diff
@@ -0,0 +1,81 @@
...
+    val keyPair = generator.generateKeyPair
+    keyPair
+  }
+
+  private def writePemFile(key: Key, description: String, file: File): Unit = {
+    new PemFile(key, description, file).write
+  }
+}
+
+class PemFile(val key: Key, val description: String, val file: File) {
```
adamnfish: Why is this a class with a method on it? Feels like we have a datastructure (model) and a function that writes? Especially true if this function is doing I/O. If we do take this approach you'll probably find it's better if the write function takes the file as an argument.

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
```diff
@@ -0,0 +1,81 @@
...
+  private val pemObject = new PemObject(description, key.getEncoded)
+
+  def write(): Unit = {
+    val pemWriter = new PemWriter(new OutputStreamWriter(new FileOutputStream(file)))
+    try
+      pemWriter.writeObject(this.pemObject)
+    finally pemWriter.close()
+    println("Wrote key to file (" + file.getCanonicalFile.toString) + ")"
```
adamnfish: We're not generally using `println`. If something needs to be said to the user then that's for the UI to do.

Also, given the try/finally, is this something that could fail?

src/main/scala/com/gu/ssm/utils/chmod.scala
```diff
@@ -0,0 +1,81 @@
+package com.gu.ssm.utils
```
adamnfish: All the fns in here feel like the same functionality to me "file system manipulations". Can we make this one Object with that purpose and an appropriate name?

src/main/scala/com/gu/ssm/utils/chmod.scala
```diff
@@ -0,0 +1,81 @@
...
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.attribute.{PosixFilePermission, PosixFilePermissions}
+
+import scala.util.Try
+
+/**
+  * Setting the file permissions
+  */
+object chmod {
```
adamnfish: Objects need capital letter names. This is Scala style (which would be enough by itself) but it's important because match statements treat capitals as types and lower case names as variables to be bound. (I realise you won't be matching on these objects so not relevant, but that's one of the reasons why that rules exists).

src/main/scala/com/gu/ssm/utils/chmod.scala
```diff
@@ -0,0 +1,81 @@
...
+      sys.error("Error setting permissions " + perms + " on " + file.getAbsolutePath + ": " + e.getMessage)
+    }
+  }
+}
+
+/**
+  * Converts a octal unix permission representation into
+  * a java `PosiFilePermissions` compatible string.
+  */
+object permissions {
```
adamnfish: Capital letter again, or (better) a single object per file (in this case we have unixy posixy behaviour. In particular, this file is called chmod.scala, and has behaviour in it that doesn't relate to chmod.

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.scalatest.matchers.BeMatcher
+
+
+class SSHTest extends FreeSpec with Matchers with EitherValues {
+  "create add key command" - {
+    import SSH.addKeyCommand
+
+    "correct content" in {
+      addKeyCommand("XXX") should include ("/bin/mkdir -p /home/ubuntu/.ssh;")
+        addKeyCommand("XXX") should include ("/bin/echo 'XXX' >> /home/ubuntu/.ssh/authorized_keys;")
```
adamnfish: Indentation is a bit off here

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+import com.gu.ssm.utils.attempt.FailedAttempt
+import org.scalatest.matchers.BeMatcher
+
+
+class SSHTest extends FreeSpec with Matchers with EitherValues {
+  "create add key command" - {
+    import SSH.addKeyCommand
+
+    "correct content" in {
```
adamnfish: Better to have one assertion per test wherever possible. This could be three separate tests (i.e. three in blocks at the same level of nesting, each with a message that explains what we're looking for).

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -0,0 +1,59 @@
...
+      addKeyCommand("XXX") should include ("/bin/mkdir -p /home/ubuntu/.ssh;")
+        addKeyCommand("XXX") should include ("/bin/echo 'XXX' >> /home/ubuntu/.ssh/authorized_keys;")
+        addKeyCommand("XXX") should include ("/bin/chmod 0600 /home/ubuntu/.ssh/authorized_keys;")
+    }
+
+  }
+
+  "create taintedcommand" - {
+
+    "create tainted command" in {
```
adamnfish: Same here, better to test each of these things independently.

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -0,0 +1,59 @@
...
+      cmd should include ("[[ -f /etc/update-motd.d/99-tainted ]] || /bin/echo -e '#!/bin/bash' | /usr/bin/sudo /usr/bin/tee -a /etc/update-motd.d/99-tainted >> /dev/null;")
+      cmd should include ("This instance was tainted by XXX at") // much text removed from this because of color codes
+      cmd should include ("/bin/chmod 0755 /etc/update-motd.d/99-tainted;")
+      cmd should include ("/usr/bin/sudo /bin/run-parts /etc/update-motd.d/ | /usr/bin/sudo /usr/bin/tee /run/motd.dynamic >> /dev/null;")
+    }
+  }
+
+  "create ssh command" - {
+    import SSH.sshCmd
+    import scala.concurrent.ExecutionContext.Implicits.global
```
adamnfish: This one should probs go at the top of the file.

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    "create ssh command" in {
+      val file:File = new File("/banana")
```
adamnfish: Probs don't need this type annotation

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -0,0 +1,70 @@
+package com.gu.ssm
+
+import java.io.File
+
+import com.amazonaws.regions.{Region, Regions}
+import scopt.OptionParser
+
+trait ArgumentParser {
```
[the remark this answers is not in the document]
aug24: Done - now an import.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -1,122 +1,81 @@
...
 
   def main(args: Array[String]): Unit = {
-    // TODO attempt to read from stdin to get commands and populate initial arguments thus
-
     argParser.parse(args, Arguments.empty()) match {
-      case Some(Arguments(Some(executionTarget), _, Some(profile), region, interactive)) if interactive =>
-        val stsClient = STS.client(profile, region)
-        val ssmClient = SSM.client(profile, region)
-        val ec2Client = EC2.client(profile, region)
+      case Some(Arguments(Some(executionTarget), toExecute, Some(profile), region, interactive, ssh)) => {
```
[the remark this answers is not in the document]
aug24: Done.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -1,122 +1,81 @@
...
+        implicit val stsClient = STS.client(profile, region)
```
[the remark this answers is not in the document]
aug24: I'm going to leave this for now...  we can reconsider when the bulk is merged.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -1,122 +1,81 @@
...
+        implicit val ec2Client = EC2.client(profile, region)
 
-        // resolve user and instances in parallel
-        val configAttempt = Attempt.map2(
-          IO.resolveInstances(executionTarget, ec2Client),
-          STS.getCallerIdentity(stsClient)
-        )((_, _))
-        val interactive = new InteractiveProgram(ssmClient)(ec)
-        interactive.main(configAttempt)
+        (toExecute, interactive, ssh) match {
```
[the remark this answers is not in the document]
aug24: I'm going to leave this for now... we can reconsider when the bulk is merged.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -1,122 +1,81 @@
...
-          config <- Attempt.map2(IO.resolveInstances(executionTarget, ec2Client), STS.getCallerIdentity(stsClient))((_, _))
-          (instances, name) = config
-          cmd <- Attempt.fromEither(Logic.generateScript(toExecute))
-          results <- IO.executeOnInstances(instances, name, cmd, ssmClient)
-        } yield results
-        val programResult = Await.result(fProgramResult.asFuture, 25.seconds)
+  private def setUpSSH(executionTarget: ExecutionTarget)(implicit stsClient: AWSSecurityTokenServiceAsync, ssmClient: AWSSimpleSystemsManagementAsync, ec2Client: AmazonEC2Async) = {
+    val fProgramResult = for {
+      config <- Attempt.map2(IO.resolveInstances(executionTarget, ec2Client), STS.getCallerIdentity(stsClient))((_, _))
+      (instances, name) = config
```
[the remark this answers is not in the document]
aug24: I'm going to leave this for now... we can reconsider when the bulk is merged.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+import java.security.{NoSuchAlgorithmException, NoSuchProviderException}
+import java.util.Calendar
+
+import scala.concurrent.ExecutionContext
+import com.gu.ssm.utils.attempt._
+import com.gu.ssm.utils.{KeyMaker, chmod}
+
+object SSH {
+
+  val delay = 30
```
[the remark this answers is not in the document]
aug24: Now "sshCredentialsLifetimeSeconds"

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def createKey(): Either[FailedAttempt, (File, String)] = {
+
+    // Write key to file.
+    val prefix = "security-magic-rsa-private-key"
```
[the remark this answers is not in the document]
aug24: Now "security_ssm-scala_temporary-rsa-private-key"

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -0,0 +1,76 @@
...
+      val tempFile = File.createTempFile(prefix, suffix)
+      chmod(tempFile, "0600")
+      val authKey = KeyMaker.makeKey(tempFile)
+      Right((tempFile, authKey))
+    } catch {
+      case e:IOException => Left(FailedAttempt(
+        Failure(s"Unable to create private key file", "Error creating key on disk", UnhandledError, None, Some(e))
+      ))
+      case e:NoSuchAlgorithmException => Left(FailedAttempt(
+        Failure(s"Unable to create key pair with that algorithm", "Error creating key", UnhandledError, None, Some(e))
```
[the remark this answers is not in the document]
aug24: Done.

src/main/scala/com/gu/ssm/SSH.scala
aug24: Done.

"This instance should be considered tainted.  It was accessed by XXX at..."

src/main/scala/com/gu/ssm/SSH.scala
aug24: Moved to Main.  There might yet be a better place...

src/main/scala/com/gu/ssm/SSH.scala
aug24: Removed entirely.

src/main/scala/com/gu/ssm/aws/EC2.scala
aug24: Didn't know that :)   Done.

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
aug24: 'println' shouldn't have been there.

In theory it can fail, and the finally guarantees the close.  I don't think there's much mileage in it though, if your disk is full/crashing, then you have bigger problems than an open file handle hanging around.

So: refactored to a 'normal' procedural approach.

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
aug24: Gone.

src/main/scala/com/gu/ssm/utils/chmod.scala
aug24: Refactored.

src/main/scala/com/gu/ssm/utils/chmod.scala
aug24: Refactored.

src/test/scala/com/gu/ssm/SSHTest.scala
aug24: Done.

src/test/scala/com/gu/ssm/SSHTest.scala
aug24: Done.

src/test/scala/com/gu/ssm/SSHTest.scala
aug24: Left as is for now, as discussed.

src/test/scala/com/gu/ssm/SSHTest.scala
aug24: Done.

src/main/scala/com/gu/ssm/IO.scala
```diff
@@ -3,29 +3,45 @@ package com.gu.ssm
...
+import com.gu.ssm.utils.attempt.{ArgumentsError, Attempt, FailedAttempt, Failure}
 
 import scala.concurrent.ExecutionContext
 
 
 object IO {
   def resolveInstances(executionTarget: ExecutionTarget, ec2Client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[List[Instance]] = {
-    executionTarget.instances.map(Attempt.Right).orElse {
+    executionTarget.instances.map( instances =>
+      EC2.resolveInstanceIds(instances, ec2Client)
```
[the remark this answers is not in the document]
aug24: Not currently addressed.  Need to have a think about exactly what we should do and where we should do it.

src/main/scala/com/gu/ssm/utils/KeyMaker.scala
aug24: Done.

src/main/scala/com/gu/ssm/IO.scala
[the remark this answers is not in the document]
aug24: Added to Trello

src/main/scala/com/gu/ssm/aws/EC2.scala
aug24: Added to Trello

README.md
```diff
@@ -26,7 +26,7 @@ You can then write a simple wrapper script and put it on your path:
 
 Call it `ssm` and make sure it is executable.
 
-## Arguments
+# Arguments
```
adamnfish: Why would this be an H1? It's not the title of the document!

README.md
```diff
@@ -51,7 +51,9 @@ specifying App, Stack, and Stage tags.
 If you provide tags, `ssm` will search for running instances that are
 have those tags.
 
-## Commands
+# Actions
```
adamnfish: Ditto, h1 is not correct here

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -6,39 +6,45 @@ import java.util.Calendar
 
 import scala.concurrent.ExecutionContext
 import com.gu.ssm.utils.attempt._
-import com.gu.ssm.utils.{KeyMaker, chmod}
+import com.gu.ssm.utils.{KeyMaker, FilePermissions}
 
 object SSH {
 
-  val delay = 30
+  val sshCredentialsLifetimeSeconds = 30
```
adamnfish: Much nicer :-)

PR #7, opened 2018-01-23
Add taint tag to instances after temporary ssh

aug24:


src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,9 +45,9 @@ object Main {
       (instances, name) = config
       sshArtifacts <- Attempt.fromEither(SSH.createKey())
       (authFile, authKey) = sshArtifacts
-      addAndRemoveKeyCommand = SSH.addTaintedCommand(name) +SSH.addKeyCommand(authKey) + SSH.removeKeyCommand(authKey)
+      addAndRemoveKeyCommand = SSH.addTaintedCommand(name) + SSH.addKeyCommand(authKey) + SSH.removeKeyCommand(authKey)
       instance <- Attempt.fromEither(getSingleInstance(instances))
-      _ <- IO.fireAndForgetOnInstances(instance, name, addAndRemoveKeyCommand, ssmClient)
+      _ <- IO.tagAndInstallSshKey(instance, name, addAndRemoveKeyCommand, ssmClient, ec2Client)
```
adamnfish: I feel like tagging and sending the SSH command should be separate explicit steps. Also, the safer way round would be to tag the instances first.

README.md
```diff
@@ -133,3 +133,6 @@ created jar file, but it's likely to be in `target/scala-2.12/ssm.jar`
 
 The jar can then be invoked as follows (it may be useful to create a
 wrapper script, as described above in 'Installation').
+
+    java -jar <path-to-jar>/ssm.jar "@$"
```
adamnfish: Should be `$@` (this was my typo from before)

src/main/scala/com/gu/ssm/aws/EC2.scala
```diff
@@ -47,4 +47,14 @@ object EC2 {
       instanceId = awsInstance.getInstanceId
     } yield Instance(InstanceId(awsInstance.getInstanceId), Option(awsInstance.getPublicIpAddress))).toList
   }
+
+  def tagInstances(ids:List[InstanceId], key: String, value: String, client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[Unit] = {
+    val request = new CreateTagsRequest()
+      .withTags(new Tag(key, value))
+      .withResources(ids.map(i => i.id).asJava)
```
adamnfish: `_.id` would be a little simpler here (very minor point)

README.md
aug24: Done

src/main/scala/com/gu/ssm/Main.scala
aug24: I explicitly changed that, having originally written it as two steps.

I'll change it back then.

aug24: It does tag first, btw.

src/main/scala/com/gu/ssm/aws/EC2.scala
aug24: Done

PR #9, opened 2018-01-24
Convert option parsing to use commands for the three major choices

katebee:
#### 👉 Switch to 'mode' selection and use commands to help with option parsing:

`ssm ssh -t app,stack,stage --profile awsprofile`

`ssm repl -t app,stack,stage --profile awsprofile`

`ssm cmd -c ls -t app,stack,stage --profile awsprofile`

picture 164

#### 👉 Separate ssh UI
Separate UI with placeholder output when running the ssh command

#### 👉 Reduce the work and usage of generateScript
convert any cmd file from within the argParser, and just pass a string around from then on

#### 👉 Delete some checks on the config
the strict requirements on commands and some opts (profile) now make them unnecessary

#### 👉 Other minor fixes
Update the tests, delete unneeded methods, spelling, formatting, etc.

### DIFF
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -1,15 +1,15 @@
 package com.gu.ssm
 
-import com.gu.ssm.utils.attempt.{ArgumentsError, FailedAttempt, Failure}
+import java.io.File
 
 import scala.io.Source
 
 
 object Logic {
-  def generateScript(toExecute: ToExecute): Either[FailedAttempt, String] = {
-    val scriptSourceOpt = toExecute.scriptOpt.map(Source.fromFile(_, "UTF-8").mkString)
-    toExecute.cmdOpt.orElse(scriptSourceOpt).toRight {
-      Failure("No execution commands provided", "You must provide commands to execute (src-file or cmd)", ArgumentsError).attempt
+  def generateScript(toExecute: Either[String, File]): String = {
+    toExecute match {
+      case Right(script) => Source.fromFile(script, "UTF-8").mkString
+      case Left(cmd) => cmd
     }
   }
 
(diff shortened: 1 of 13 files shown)


src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -1,15 +1,15 @@
...
 
 object Logic {
-  def generateScript(toExecute: ToExecute): Either[FailedAttempt, String] = {
-    val scriptSourceOpt = toExecute.scriptOpt.map(Source.fromFile(_, "UTF-8").mkString)
-    toExecute.cmdOpt.orElse(scriptSourceOpt).toRight {
-      Failure("No execution commands provided", "You must provide commands to execute (src-file or cmd)", ArgumentsError).attempt
+  def generateScript(toExecute: Either[String, File]): String = {
+    toExecute match {
+      case Right(script) => Source.fromFile(script, "UTF-8").mkString
+      case Left(cmd) => cmd
```
katebee: Could arguably simplify this further, and make the argument a String, since it is only used to parse any file when running in cmd mode... Could even delete it?

However, this PR was getting out of control so I wanted to save some changes for another day!

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -15,17 +17,18 @@ object Main {
 
   def main(args: Array[String]): Unit = {
     argParser.parse(args, Arguments.empty()) match {
-      case Some(Arguments(Some(executionTarget), toExecuteOpt, Some(profile), region, interactive, ssh)) =>
+      case Some(Arguments(Some(executionTarget), toExecuteOpt, Some(profile), region, mode)) =>
```
guardian-data-security: Wouldn't this be nicer as Some(mode)?  And if not, then the variable would probably be clearer as modeOpt.

shtukas:
Note to self: when doing the `ssh` command the interactive session colours need to be corrected.

PR #11, opened 2018-01-25
Alignment change in the ssh command user interface

shtukas:


shtukas:
Before:

screen shot 2018-01-25 at 14 24 36

shtukas:
After:

screen shot 2018-01-25 at 14 27 29

adamnfish:
I think the leading space was there on purpose? A leading space prevents the command from going into your bash history. In this case it's a temporary one-shot command so there's no reason you'd want it to be present after you've run it.

Note: this is also what Janus does to prevent pasted credentials from showing up.

shtukas:
Point taken. I have (re) discovered something today. Closing

aug24:
+1

PR #12, opened 2018-01-25
Comes back to default colour instead of hard coded black, after red.

shtukas:
Comes back to default colour instead of hard coded black, after red, during ssm command


shtukas:
screen shot 2018-01-25 at 14 50 42

aug24:
+1

PR #13, opened 2018-01-30
UI Improvment: reporting erroneous instance ids while using cmd

shtukas:
With this change while running

```
$ java -jar ./ssm.jar cmd -c ls --profile security --instances i-011a0282bac767f41,i-09d626a5abdfb5c74,i-123456
```

where only the fist instance id (`i-011a0282bac767f41`) is correct, the output now becomes

```
The following instance(s) could not be found: i-09d626a5abdfb5c74, i-123456

========= i-011a0282bac767f41 =========
STDOUT:
amazon-ssm-agent.deb
bin
(...)
vmlinuz

STDERR:

```

### DIFF
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -52,15 +52,12 @@ object Main {
     System.exit(programResult.fold(_.exitCode, _ => 0))
   }
 
-  private[ssm] def computeIncorrectInstances(executionTarget: ExecutionTarget, results: List[(InstanceId, scala.Either[CommandStatus, CommandResult])]): List[InstanceId] =
-    executionTarget.instances.getOrElse(List()).filterNot(results.map(_._1).toSet)
-
   private def execute(awsClients: AWSClients, profile: String, region: Region, executionTarget: ExecutionTarget, toExecute: String): Unit = {
     val fProgramResult = for {
       config <- IO.getSSMConfig(awsClients.ec2Client, awsClients.stsClient, profile, region, executionTarget)
       _ <- Attempt.fromEither(Logic.checkInstancesList(config))
       results <- IO.executeOnInstances(config.targets.map(i => i.id), config.name, toExecute, awsClients.ssmClient)
-      incorrectInstancesFromInstancesTag = computeIncorrectInstances(executionTarget, results)
+      incorrectInstancesFromInstancesTag = Logic.computeIncorrectInstances(executionTarget, results)
     } yield ResultsWithInstancesNotFound(results,incorrectInstancesFromInstancesTag)
     val programResult = Await.result(fProgramResult.asFuture, maximumWaitTime)
     programResult.fold(UI.outputFailure, UI.output)
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -75,6 +75,12 @@ object Main {
     val programResult = Await.result(fProgramResult.asFuture, 25.seconds)
 
     programResult.fold(UI.outputFailure, UI.output)
+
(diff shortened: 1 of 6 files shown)


src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -75,6 +75,12 @@ object Main {
     val programResult = Await.result(fProgramResult.asFuture, 25.seconds)
 
     programResult.fold(UI.outputFailure, UI.output)
+
+    val leftOverInstances = executionTarget.instances.getOrElse(List()).filterNot(programResult.right.get.map(x => x._1).toSet)
```
adamnfish: This is neat.

Consider pulling the logic out of Main so it can be tested, though?

adamnfish: (would also let the tested logic be re-used in your follow-up REPL PR)

shtukas: Good idea. But if you don't mind, I will do the refactoring affecting the two change in a separate PR.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -59,6 +59,11 @@ object Main {
       results <- IO.executeOnInstances(config.targets.map(i => i.id), config.name, toExecute, awsClients.ssmClient)
     } yield results
     val programResult = Await.result(fProgramResult.asFuture, maximumWaitTime)
+    val leftOverInstances = executionTarget.instances.getOrElse(List()).filterNot(programResult.right.get.map(x => x._1).toSet)
```
adamnfish: The await should be the last thing that happens, let's talk about why this happens after our program should have "finished"?

shtukas:
@adamnfish 

I think you are going to like my last update much better. It should actually be self explanatory but just in case...

`case class ResultsWithInstancesNotFound` does exactly what is says on the tin [1]. The `UI.output` function has been extended in the most natural way. 

Now, the thing is that in the `main`'s `execute` function all I am doing, so far is reporting `incorrectInstancesFromInstancesTag`. In an ideal world I should be able to do better than that (as we discussed, as discussed it would be nice to also do the **--tags** argument), but for the moment, this will do. The important thing is that the new case class and the UI update won't need to be updated :)

[1] I am happy to be told that this case class has a better name :)

src/test/scala/com/gu/ssm/MainTest.scala
```diff
@@ -1,8 +1,22 @@
 package com.gu.ssm
 
 import org.scalatest.{EitherValues, FreeSpec, Matchers}
+import com.gu.ssm.Main.computeIncorrectInstances
 
 
 class MainTest extends FreeSpec with Matchers with EitherValues {
 
+  "computeIncorrectInstances" - {
+    "compute correct difference 1" in {
```
adamnfish: Be nice if these were a little more descriptive. Looks like this one is expecting to find no incorrect instances and the other should return something incorrect?

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -52,12 +52,16 @@ object Main {
     System.exit(programResult.fold(_.exitCode, _ => 0))
   }
 
+  private[ssm] def computeIncorrectInstances(executionTarget: ExecutionTarget, results: List[(InstanceId, scala.Either[CommandStatus, CommandResult])]): List[InstanceId] =
```
adamnfish: It doesn't look like this belongs in Main?

src/test/scala/com/gu/ssm/MainTest.scala
shtukas: Done! (in change "Better test names")

src/main/scala/com/gu/ssm/Main.scala
shtukas: Done! In change called "refactoring computeIncorrectInstances away from Main"

PR #14, opened 2018-01-30
UI Improvment: reporting erroneous instance ids while using repl

shtukas:
We now report in the repl screen that some instances could not be found. 
This is the sister PR of: https://github.com/guardian/ssm-scala/pull/13

Running 

```
$ java -jar ./ssm.jar repl --profile security --instances i-04d908c9a645cd264,i-123456
```

where the only valid id is `i-04d908c9a645cd264`, we get

screen shot 2018-01-30 at 14 00 13

The following screenshot shows the error message shown after the command having been ran ( `i-049447ce9628d26fe` was correct instance id and `i-123456` an incorrect one )

screen shot 2018-01-30 at 15 11 27

This change contains few function signatures changes to carry `executionTarget` where it wasn't originally available.

### DIFF
--- a/src/main/scala/com/gu/ssm/Interactive.scala
+++ b/src/main/scala/com/gu/ssm/Interactive.scala
@@ -25,24 +25,24 @@ class InteractiveProgram(val awsClients: AWSClients)(implicit ec: ExecutionConte
       _ <- Attempt.fromEither(Logic.checkInstancesList(config))
     } yield config
 
-    val userSubmittedInstanceIds = executionTarget.instances.getOrElse(Nil)
-
     configAttempt.onComplete {
-      case Right(SSMConfig(targets, name)) =>
-        ui.ready(targets.map(i => i.id), name, userSubmittedInstanceIds)
+      case Right(SSMConfig(targets, name)) => {
+        val incorrectInstancesFromInstancesTag = executionTarget.instances.getOrElse(Nil).filterNot(targets.map(i => i.id).toSet)
+        ui.ready(targets.map(i => i.id), name, incorrectInstancesFromInstancesTag)
+      }
       case Left(failedAttempt) =>
         ui.displayError(failedAttempt)
-        ui.ready(List(), "", userSubmittedInstanceIds)
+        ui.ready(List(), "", Nil)
     }
   }
 
   /**
     * Kick off execution of a new command and update UI when it returns
     */
-  def executeCommand(command: String, instances: List[InstanceId], username: String, userSubmittedInstanceIds: List[InstanceId]): Unit = {
+  def executeCommand(command: String, instances: List[InstanceId], username: String, instancesNotFound: List[InstanceId]): Unit = {
     IO.executeOnInstances(instances, username, command, awsClients.ssmClient).onComplete {
       case Right(results) =>
-        ui.displayResults(instances, username, results, userSubmittedInstanceIds)
+        ui.displayResults(instances, username, ResultsWithInstancesNotFound(results, instancesNotFound))
       case Left(fa) =>
         ui.displayError(fa)
     }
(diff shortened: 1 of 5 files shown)


src/main/scala/com/gu/ssm/Interactive.scala
```diff
@@ -57,7 +58,7 @@ class InteractiveUI(program: InteractiveProgram) extends LazyLogging {
   /**
     * Create window that displays the main UI along with the results of the previous command
     */
-  def mainWindow(instances: List[InstanceId], username: String, results: List[(InstanceId, Either[CommandStatus, CommandResult])]): BasicWindow = {
+  def mainWindow(executionTarget: ExecutionTarget, instances: List[InstanceId], username: String, results: List[(InstanceId, Either[CommandStatus, CommandResult])]): BasicWindow = {
```
adamnfish: Why do we need execution target *and* instances here?

shtukas: They are not the same. Instances where the instances AWS could find, not the instances provided by AWS.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -84,6 +84,6 @@ object Main {
       STS.getCallerIdentity(stsClient)
     )((_, _))
     val interactive = new InteractiveProgram(ssmClient)(ec)
-    interactive.main(configAttempt)
+    interactive.main(executionTarget,configAttempt)
```
adamnfish: instances

shtukas: Same reply as above. They are not the same....

shtukas: `executionTarget` carries all the instance Ids provided by the user, but not `instances` at this stage.

shtukas: [edit: deleted my two previous answers]

Actually `instances` is not in scope here.

src/main/scala/com/gu/ssm/Interactive.scala
shtukas: They are not the same. `instances` contains the instances AWS could find, not the instances provided by the user.

adamnfish:
I think it's a bit odd that the message saying "not found" is put at the bottom of the output for a different instance. The UI here is:

[ input ]

[ instance selector dropdown \/ ]

[ instance id ]
.. output for the above instance

In the first case, if we know that the provided instances are invalid then arguably we should fail straight away. In the second we don't discover the problem until after we've attempted to execute a command. In this case the approach makes sense, but maybe better to put the instance information along with the instance output selector. What should happen to the instance output selector in this case?

I do think the `executionTarget` type probably isn't relevant to the REPL UI so it's a bit confusing the way we're passing what looks liek duplicate info about at the moment. Sadly, I have to run to a few hours of meetings now, so have a think and I'll be able to take another look later.

shtukas:
@adamnfish 

> I have to run to a few hours of meetings now, so have a think and I'll be able to take another look later.

No problem. Take your time. I am already doing something else anyway, so this is not a blocker :)

shtukas:
@adamnfish 

> I think it's a bit odd that the message saying "not found" is put at the bottom of the output for a different instance.

Actually that's not the way it was intended to be read. You read it like that due to lack of real visual design (my bad), but that error is independent from the output of any particular instance. Let me see if I can do something about that...

shtukas:
@adamnfish 

The following picture (I have updated the commit message), should address the layout confusion.

screen shot 2018-01-30 at 15 11 27

shtukas:
@adamnfish 

> In the first case, if we know that the provided instances are invalid then arguably we should fail straight away.

This is debatable. For the record, I fundamentally totally agree with you, but the purpose of my change was to highlight when an instance id wasn't valid, which the change does. The change's purpose wasn't to modify the existing execution logic. I am more than happy to do that, but we need to agree on what the (new) logic should be :)

adamnfish:
UI is much better like that, nice one.

Having an argument of type `ExecutionTarget` with name `executionTarget` that has nothing to do with the target of execution is not what we're after still though, I think. Could it not be a list of the instances that were not found, and preferably the last argument rather than the first (to show that it is ancillary)?

shtukas:
> Could it not be a list of the instances that were not found, and preferably the last argument rather than the first (to show that it is ancillary)?

Sure. You are right, Let me do that...

shtukas:
@adamnfish 

The refactoring here works along the same principles as here: https://github.com/guardian/ssm-scala/pull/13#issuecomment-362423365

Same `ResultsWithInstancesNotFound` case class, and same computation of `incorrectInstancesFromInstancesTag` in the `main` function.

PR #15, opened 2018-01-30
Improve config data structure

aug24:


src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -61,29 +70,38 @@ object Main {
   }
 
   def getSingleInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
-    if (instances.length != 1) Left(FailedAttempt(
+    if (instances.lengthCompare(1) != 0) Left(FailedAttempt(
```
adamnfish: :-/

Your way is better than IntelliJ's here, I reckon.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -61,29 +70,38 @@ object Main {
...
 
-  private def interactiveLoop(executionTarget: ExecutionTarget)(implicit stsClient: AWSSecurityTokenServiceAsync, ssmClient: AWSSimpleSystemsManagementAsync, ec2Client: AmazonEC2Async): Unit = {
-    val configAttempt = Attempt.map2(
-      IO.resolveInstances(executionTarget, ec2Client),
-      STS.getCallerIdentity(stsClient)
-    )((_, _))
-    val interactive = new InteractiveProgram(ssmClient)(ec)
-    interactive.main(configAttempt)
+  private def readEvaluatePrintLoop(ssmConfig: Attempt[SSMConfig]): Unit = {
+    Await.result(ssmConfig.asFuture, 25.seconds) match {
```
adamnfish: This shouldn't block. The REPL starts straight away and waits for the config to resolve (you'll see it accepts an Attempt)

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -61,29 +70,38 @@ object Main {
...
+      case Right(config) =>
+        new InteractiveProgram(config)(ec)
+      case Left(failedAttempt) =>
+        UI.outputFailure(failedAttempt)
+        System.exit(failedAttempt.exitCode)
+    }
   }
+
+  case class SSMConfig (
```
adamnfish: This should be in models, and be formatted in one of these ways (depending on length). It's silly that IntelliJ is misconfigured by default, remind me to look up the setting for you when you're around.

```scala
case class Name(arg1: Type, arg2: Type)

case class Name(
  arg1: Type,
  arg2: Type,
  arg3: Type,
  arg4: Type,
  arg5: Type,
  arg6: Type,
  arg7: Type,
)

File: src/main/scala/com/gu/ssm/Interactive.scala [hunk 4]
```diff
@@ -1,40 +1,33 @@
...
-      ui.start()
-    }
-    // update UI when we're ready to get started
-    setupAttempt.onComplete {
-      case Right((instances, username)) =>
-        ui.ready(instances.map(i => i.id), username)
-      case Left(fa) =>
-        ui.displayError(fa)
-    }
+  // start UI on a new thread (it blocks while it listens for keyboard input)
```
adamnfish: I've a mild preference for not having the side-effect on the constructor, but I could maybe be convinced away from that. Either way, the program is designed to start immediately and wait for the config Attempt to complete. We should preserve that behaviour here, I think.

src/main/scala/com/gu/ssm/Interactive.scala
```diff
@@ -1,40 +1,33 @@
...
+  Future {
+    ui.start()
   }
+  // update UI when we're ready to get started
```
adamnfish: This PR means the program blocks for up to 25 seconds before it even gets to here, which is probably not what is intended? I've mentioned elsewhere, this should still take the attempt as a parameter. With the code written thus, it becomes a little clearer why this happened from a main function rather than the constructor (since it is expecting events to happen and kicks off a bunch of IO).

src/main/scala/com/gu/ssm/Main.scala
aug24: Done.

PR #21, opened 2018-02-09
Simplify use by providing an executable

shtukas:
This change introduces `generate-executable.sh` which can now be used to generate single executable called `ssm`. The documentation has been updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,16 +10,7 @@ Instructions for using SSM Scala in your own project can be found [below](#How-t
 
 ## Installation
 
-Fetch the most recently released version of the program from the
-Github releases page
-
-[Permalink to latest release](https://github.com/guardian/ssm-scala/releases/latest)
-
-You can then write a simple wrapper script and put it on your path:
-
-    java -jar <path-to-jar>/ssm.jar "$@"
-
-Call it `ssm` and make sure it is executable (`chmod +x ssm`).
+Fetch the most recently released version of the program from the [Github releases page](https://github.com/guardian/ssm-scala/releases/latest) and make sure it is executable (`chmod +x ssm`).
 
 ## Usage
 
@@ -47,19 +38,13 @@ create and upload a temporary ssh key
 The general syntax is 
 
 ```
-ssm [cmd|repl|ssh] [options]
+./ssm [cmd|repl|ssh] [options]
 ```
 
-An example of `cmd` (from the same folder where `ssm.jar` is located) is 
+An example of `cmd` is 
 
 ```
-java -jar ./ssm.jar cmd -c ls --profile security -t security-hq,security,PROD
-```
-
-... but, as per advised in the **Installation** section, if you have the `ssm` wrapper set up this can be rewritten 
-
-```
-ssm cmd -c ls --profile security -t security-hq,security,PROD
+./ssm cmd -c ls --profile security -t security-hq,security,PROD
 ```
 
 For more information about `--profile` see [AWS profiles](https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -149,20 +134,17 @@ sbt shell or from the CLI in that project.
     sbt:ssm-scala> run --instances i-0123456 --profile xxx --region xxx --cmd pwd
 
 However, `sbt` traps the program exit so in REPL mode you may find it
-easier to create and run a jar instead.
-
-You can produce a `jar` from your current source code using `assembly`
-from the sbt shell, or by executing the following from within the
-project. The command's output will show the location of the newly
-created jar file, but it's likely to be in `target/scala-2.12/ssm.jar`
+easier to create and run a an executable instead, for this just run
```
katebee: > a an

🙈 sorry!

shtukas: Done, thanks!

generate-executable.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+sbt assembly
+cat generate-executable-prefix target/scala-2.12/ssm.jar > target/scala-2.12/ssm
```
adamnfish: This assumes:
* the script will be executed from the correct location
* that the relative path will be `target/scala-2.12/ssm.jar`

The latter assumption is borderline. It'll work now, but we'll definitely update the Scala version in the future and this will definitely break when that happens.

The former is probably not a safe assumption. Someone will click download, then run `~/Downloads/generate-executable.sh` and this will break. You can do bash guff to get the absolute path from the script's location.

generate-executable-prefix
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
adamnfish: This is fascinating. So Java will happily skip the script prefix and go straight onto reading the embedded binary?

README.md
```diff
@@ -10,16 +10,7 @@ Instructions for using SSM Scala in your own project can be found [below](#How-t
...
-Github releases page
-
-[Permalink to latest release](https://github.com/guardian/ssm-scala/releases/latest)
-
-You can then write a simple wrapper script and put it on your path:
-
-    java -jar <path-to-jar>/ssm.jar "$@"
-
-Call it `ssm` and make sure it is executable (`chmod +x ssm`).
+Fetch the most recently released version of the program from the [Github releases page](https://github.com/guardian/ssm-scala/releases/latest) and make sure it is executable (`chmod +x ssm`).
```
adamnfish: I think we should still say words to the effect of "optionally, put it on your path"

rtyley:
Well done on figuring that out! Looks like a good reference on this technique is:

https://skife.org/java/unix/2011/06/20/really_executable_jars.html

I think that's where the Presto project got the idea from.

shtukas:
@rtyley :+1:

generate-executable-prefix
shtukas: Yep; and from me to you, this is exactly want I needed to know to move from Ruby to Scala as my home "scripting" language :)

generate-executable.sh
shtukas: About the scala version, you are right. My thinking was that if we change the it, then  it will be obvious that this needs to be updated. I understand this is not ideal.

shtukas: > the script will be executed from the correct location

Yes. I guess I can add "From the root of the repository...."

shtukas: ... but I will make it more robust :)

shtukas: @adamnfish Done!

generate-executable.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $DIR
+sbt assembly
+cat "$DIR/generate-executable-prefix" "$DIR/target/scala-2.12/ssm.jar" > "$DIR/target/scala-2.12/ssm"
```
adamnfish: Further to our chat earlier, is it worth checking if the file exists at `$DIR/target/scala-2.12/ssm.jar` and echoing a message saying "it looks like this file has moved (maybe the Scala version changed). Check the output of `sbt assembly` and update the script accordingly". Not urgent, I'm just thinking about staving off confusion in the future when someone that isn't us works on this.

shtukas: Good idea, will do :)

PR #22, opened 2018-02-12
New (optional) command line argument in the case of the ssh command: -a, --any <boolean>

shtukas:
- Some minor stylistic updates in arguments documentation.
- Added a new (optional) command line argument in the case of the ssh command.
- Updated the code to process this extra argument.

The net result is that users can now specify that they want ssh to proceed with one single instance if more than one instances where found. (Request from the CAPI team).

```
$ ./ssm ssh -p security -t security-hq,security,PROD
Error choosing single instance, found i-09bab0e2ecef2b184, i-0792c00571352d0e4

$ ./ssm ssh -p security -t security-hq,security,PROD -a true
========= i-09bab0e2ecef2b184 =========
STDOUT:

 # Execute the following command within the next 30 seconds:
 ssh -i /private/var/folders/7b/djwwh2kd4m95smw88z4hvfnr0000gn/T/security_ssm-scala_temporary-rsa-private-key4611386747952826521.tmp ubuntu@34.245.57.216;

```

- Updated the README with the relevant information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,21 +18,23 @@ The automatically generated help section for `ssm` is
 
 ```
 Usage: ssm [cmd|repl|ssh] [options]
-  -p, --profile <value>    the AWS profile name to use for authenticating this execution
-  -i, --instances <value>  specify the instance ID(s) on which the specified command(s) should execute
-  -t, --tags <value>       search for instances by tag e.g. '--tags app,stack,stage'
+
+  -p, --profile <value>    The AWS profile name to use for authenticating this execution
+  -i, --instances <value>  Specify the instance ID(s) on which the specified command(s) should execute
+  -t, --tags <value>       Search for instances by tag e.g. '--tags app,stack,stage'
   -r, --region <value>     AWS region name (defaults to eu-west-1)
 
 Command: cmd [options]
-execute a single (bash) command, or a file containing bash commands
-  -c, --cmd <value>        a bash command to execute
-  -f, --file <value>       a file containing bash commands to execute
+Execute a single (bash) command, or a file containing bash commands
+  -c, --cmd <value>        A bash command to execute
+  -f, --file <value>       A file containing bash commands to execute
 
 Command: repl
-run SSM in interactive/repl mode
+Run SSM in interactive/repl mode
 
-Command: ssh
-create and upload a temporary ssh key
+Command: ssh [options]
+Create and upload a temporary ssh key
+  -a, --any <value>        Indicates whether the command should run on any single instance
 ```
 
 The general syntax is 
@@ -75,6 +77,8 @@ key, and install the public key on a specific instance.  It will then
 output the command to `ssh` directly to that instance. 
 The instance must already have both a public IP address _and_
 appropriate security groups.
+
+Notet that if the argument `-t <app>,<stack>,<stage>` resolves to more than one instance, the command will stop with an error message. You can circumvent this behaviour and instruct `ssm` to process with one single instance with the argument `-a true`.
 
 ### More usage examples
 
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ output the command to `ssh` directly to that instance.
 The instance must already have both a public IP address _and_
 appropriate security groups.
 
-Notet that if the argument `-t <app>,<stack>,<stage>` resolves to more than one instance, the command will stop with an error message. You can circumvent this behaviour and instruct `ssm` to process with one single instance with the argument `-a true`.
+Note that if the argument `-t <app>,<stack>,<stage>` resolves to more than one instance, the command will stop with an error message. You can circumvent this behaviour and instruct `ssm` to process with one single instance with the argument `-a true`.
 
 ### More usage examples
 
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -35,9 +35,23 @@ object Logic {
   }
 
   def getSingleInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
-    if (instances.lengthCompare(1) != 0) Left(FailedAttempt(
+    if (instances.length != 1) Left(FailedAttempt(
       Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${instances.map(i => i.id.id).mkString(", ")}", UnhandledError, None, None)))
     else Right(instances.map(i => i.id))
+  }
+
+  def getFirstInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
+    if (instances.length == 0) Left(FailedAttempt(
+      Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
+    else Right(instances.map(i => i.id).take(1))
+  }
+
+  def getRelevantInstancesAsEither(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, List[InstanceId]] = {
+    if (takeAnySingleInstance) getFirstInstance(instances) else getSingleInstance(instances)
+  }
+
+  def getRelevantInstancesAsList(instances: List[Instance], takeAnySingleInstance: Boolean): List[Instance] = {
+    if (takeAnySingleInstance) instances.take(1) else instances
   }
 
   def getClients(profile: String, region: Region): AWSClients = {
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -34,24 +34,16 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSingleInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
-    if (instances.length != 1) Left(FailedAttempt(
-      Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${instances.map(i => i.id.id).mkString(", ")}", UnhandledError, None, None)))
-    else Right(instances.map(i => i.id))
-  }
-
-  def getFirstInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
-    if (instances.length == 0) Left(FailedAttempt(
-      Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
-    else Right(instances.map(i => i.id).sortWith( (i1,i2) => i1.id < i2.id ).take(1))
-  }
-
-  def getRelevantInstancesAsEither(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, List[InstanceId]] = {
-    if (takeAnySingleInstance) getFirstInstance(instances) else getSingleInstance(instances)
-  }
-
-  def getRelevantInstancesAsList(instances: List[Instance], takeAnySingleInstance: Boolean): List[Instance] = {
-    if (takeAnySingleInstance) instances.sortWith( (i1,i2) => i1.id.id < i2.id.id ).take(1) else instances
+  def getRelevantInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
+    if (instances.length == 0) {
+      Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
+    } else if (instances.length == 1) {
+      Right(instances.sortBy(_.id.id).head) // head safe as length is necessary > 0
+    } else {
+      val sortedInstances = instances.sortBy(_.id.id)
+      if (takeAnySingleInstance) Right(sortedInstances.head) // head safe as length is necessary > 0
+      else Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedInstances.mkString(", ")}", UnhandledError, None, None)))
+    }
   }
 
   def getClients(profile: String, region: Region): AWSClients = {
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -76,6 +78,8 @@ output the command to `ssh` directly to that instance.
 The instance must already have both a public IP address _and_
 appropriate security groups.
 
+Notet that if the argument `-t <app>,<stack>,<stage>` resolves to more than one instance, the command will stop with an error message. You can circumvent this behaviour and instruct `ssm` to process with one single instance with the argument `-a true`.
```
tomrf1: typo in "Note"

tomrf1:
👍  thanks @shtukas !

README.md
shtukas: Oh thanks. Corrected.

shtukas:
@adamnfish with https://github.com/guardian/ssm-scala/pull/22/commits/79297ca49a6e4c1f52e4a290c2a46d5afe4b7322 , the selection of an `Instance` and an `InstanceId` are determinist but also compatible with each other. Meaning that selecting an `Instance` and then taking its id, or converting the `Instance`s to `InstanceId`s and then select selecting one give the same result.

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -2,6 +2,8 @@ package com.gu.ssm
 
 import org.scalatest.{EitherValues, FreeSpec, Matchers}
 
+import scala.collection.immutable.Nil
```
adamnfish: This is in predef, shouldn't need to be imported

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -41,21 +43,48 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
     }
   }
 
-  "Get single instance" - {
-    import Logic.getSingleInstance
+  "getRelevantInstance" - {
+    import Logic.getRelevantInstance
+
+    "No instances" in {
```
adamnfish: Do you feel like this error message is what you'd want to see from a failing test suite?

The full message would be:

    "getRelevantInstance No instances"

It's good to describe what you're after. Perhaps something like `"fails if no instances are supplied"` for the test itself. See what you think though (here and for the others).

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -41,21 +43,48 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
+      getRelevantInstance(List(),true).isLeft shouldBe true
+    }
+
+    "Exactly one instance" in {
+      getRelevantInstance(List(Instance(InstanceId("X"), None)), true) shouldBe Right(InstanceId("X"))
```
adamnfish: What are the two separate assertions here for? It's good to aim for one assertion per test if you can, so you can describe what you're after in each case.

`FreeSpec` allows us to group tests which can be useful for creating good test messages as well as for sharing test setup. For example in this case we could do something like the following:

```scala
"getRelevantInstance" - {
  "fails when provided no instances" in {
    getRelevantInstance(List(),true).isLeft shouldBe true
  }

  "when given exactly one instance" - {
    val instances = List(Instance(InstanceId("X"), None))

    "returns that instance when asked for a single instance" in {
      getRelevantInstance(instances, false) shouldBe Right(InstanceId("X"))
    }

    "returns that instance when asked for multiple instances" in {
      getRelevantInstance(instances, true) shouldBe Right(InstanceId("X"))
    }
  }

  "when given multiple instances" - {
    val instances = List(Instance(InstanceId("X"), None), Instance(InstanceId("Y"), None)

    "fails when asked for a single instance" in {
      getRelevantInstance(instances, false).isLeft shouldBe true
    }

    "returns the first instance when asked for multiple instances" in {
      getRelevantInstance(instances, true) shouldBe Right(InstanceId("X"))
    }
  }
}
```

This is off the top of my head so I'm not saying copy paste this, just trying to demonstrate how it's possible to build up a set of related tests that are logically grouped and that give useful errors from the terminal when they fail.

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -34,10 +34,20 @@ object Logic {
...
+    }else if(instances.length == 1){
+      Right(instances.map(i => i.id).sortWith( (i1,i2) => i1.id < i2.id ).head) // head safe as length is necessary > 0
+    }else{
+      val sortedInstances = instances.map(i => i.id).sortBy(_.id)
+      if (takeAnySingleInstance) Right(sortedInstances.head) // head safe as length is necessary > 0
+      else Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedInstances.mkString(", ")}", UnhandledError, None, None)))
+    }
+  }
+
+  def getRelevantInstances(instances: List[Instance], takeAnySingleInstance: Boolean): List[Instance] = {
```
adamnfish: Is this function used?

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -41,21 +43,48 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
+      val x = Instance(InstanceId("X"), None)
+      val y = Instance(InstanceId("Y"), None)
+      getRelevantInstances(List(x), true) shouldBe List(x)
+      getRelevantInstances(List(x), false) shouldBe List(x)
     }
+
+    "Several instances" in {
+      val x = Instance(InstanceId("X"), None)
+      val y = Instance(InstanceId("Y"), None)
+      getRelevantInstances(List(x,y), true) shouldBe List(x)
```
adamnfish: `List(x,y)` -> `List(x, y)`

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -41,21 +43,48 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
     }
   }
 
-  "Get single instance" - {
-    import Logic.getSingleInstance
+  "getRelevantInstance" - {
+    import Logic.getRelevantInstance
+
+    "No instances" in {
+      getRelevantInstance(List(),true).isLeft shouldBe true
```
adamnfish: `(List(),true)` -> `(List(), true)`

src/main/scala/com/gu/ssm/aws/SSM.scala
```diff
@@ -21,11 +21,11 @@ object SSM {
       .build()
   }
 
-  def sendCommand(instances: List[InstanceId], cmd: String, username: String, client: AWSSimpleSystemsManagementAsync)(implicit ec: ExecutionContext): Attempt[String] = {
+  def sendCommand(instanceIds: List[InstanceId], cmd: String, username: String, client: AWSSimpleSystemsManagementAsync)(implicit ec: ExecutionContext): Attempt[String] = {
```
adamnfish: Better!

src/main/scala/com/gu/ssm/aws/EC2.scala
```diff
@@ -48,10 +48,10 @@ object EC2 {
     } yield Instance(InstanceId(instanceId), Option(awsInstance.getPublicIpAddress))).toList
   }
 
-  def tagInstances(ids:List[InstanceId], key: String, value: String, client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[Unit] = {
+  def tagInstance(id:InstanceId, key: String, value: String, client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[Unit] = {
```
adamnfish: `id:InstanceId` -> `id: InstanceId`

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -36,16 +36,16 @@ object Main {
...
       (authFile, authKey) = sshArtifacts
       addAndRemoveKeyCommand = SSH.addTaintedCommand(config.name) + SSH.addKeyCommand(authKey) + SSH.removeKeyCommand(authKey)
-      instance <- Attempt.fromEither(Logic.getSingleInstance(config.targets))
-      _ <- IO.tagAsTainted(instance, config.name, awsClients.ec2Client)
-      _ <- IO.installSshKey(instance, config.name, addAndRemoveKeyCommand, awsClients.ssmClient)
-    } yield config.targets.map(SSH.sshCmd(authFile, _))
+      instanceId <- Attempt.fromEither(Logic.getRelevantInstance(config.targets, takeAnySingleInstance))
+      _ <- IO.tagAsTainted(instanceId, config.name, awsClients.ec2Client)
+      _ <- IO.installSshKey(instanceId, config.name, addAndRemoveKeyCommand, awsClients.ssmClient)
+    } yield SSH.sshCmd(authFile, Instance(instanceId, None))
```
adamnfish: The fact that we have to re-wrap the ID in a dummy `Instance` container looks like a sign that one of the following is true:

* `Logic.getRelevantInstance` should return an `Instance` instead of an `InstanceId`
* `SSH.sshCmd` should take `InstanceId`

Always good to pass as little information around as possible, so the latter is probably better.

adamnfish: Scratch that, it's the former.

https://github.com/guardian/ssm-scala/blob/master/src/main/scala/com/gu/ssm/SSH.scala#L69

And in fact, this change would totally break the SSH functionality. The public IP address is required to be able to create the SSH command.

adamnfish:
(Changes requested because this breaks the SSH command at the moment by losing the instance's public IP address)

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: Good point. I was mostly trying to stick to the existing "conventions" (most of the test suite was written before me), but they are not conventions and should be broken :)

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: True. This is one of those things that my IntelliJ does thinking it's being clever :)

shtukas:
> by losing the instance's public IP address

TIL: in

```
case class Instance(id: InstanceId, publicIpAddressOpt: Option[String])
```

`publicIpAddressOpt` is the public IP address of the instance. I had never noticed/realised that before...

shtukas:
@adamnfish 

Interestingly I have just understood why my breaking change wasn't picked up by the tests. The test for `sshCmd(file, instance)` is called with a fully specified `Instance`...

```
    "create ssh command" in {
      val file: File = new File("/banana")
      val instance: Instance = Instance(InstanceId("raspberry"), Some("127.0.0.1"))
      val cmd = sshCmd(file, instance)
      cmd._1.id shouldEqual "raspberry"
      cmd._2 should include ("ssh -i /banana ubuntu@127.0.0.1")
    }
```

I could modify the test to construct the `instance` with a call to `Attempt.fromEither(Logic.getRelevantInstance(config.targets, takeAnySingleInstance))` but that would be clearly overkill and defeat the atomicity of the test. 

Anyway, I have learnt my lesson :)

shtukas:
ps: and cherry on the cake, this now reads much more naturally
```
      instance <- Attempt.fromEither(Logic.getRelevantInstance(config.targets, takeAnySingleInstance))
      _ <- IO.tagAsTainted(instance.id, config.name, awsClients.ec2Client)
      _ <- IO.installSshKey(instance.id, config.name, addAndRemoveKeyCommand, awsClients.ssmClient)
    } yield SSH.sshCmd(authFile, instance)
```

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -63,12 +60,12 @@ object SSH {
...
 
-  def sshCmd(tempFile: File, instance: Instance)(implicit ec: ExecutionContext): (InstanceId, String) = {
+  def sshCmd(tempFile: File, instance: Instance): (InstanceId, String) = {
     val cmd = s"""
       | # Execute the following command within the next $sshCredentialsLifetimeSeconds seconds:
       | ssh -i ${tempFile.getCanonicalFile.toString} ubuntu@${instance.publicIpAddressOpt.get};
       |""".stripMargin
-    instance.id -> cmd
+    (instance.id, cmd)
   }
```
shtukas: I know `instance.id -> cmd` and `(instance.id, cmd)` are the same thing. The change in meant to mirror the signature of the function.

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -34,10 +34,16 @@ object Logic {
...
   }
 
-  def getSingleInstance(instances: List[Instance]): Either[FailedAttempt, List[InstanceId]] = {
-    if (instances.lengthCompare(1) != 0) Left(FailedAttempt(
-      Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${instances.map(i => i.id.id).mkString(", ")}", UnhandledError, None, None)))
-    else Right(instances.map(i => i.id))
+  def getRelevantInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
+    if (instances.length == 0) {
+      Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
+    }else if(instances.length == 1){
```
adamnfish: `}else if(...){` -> `} else if (...) {`

adamnfish: Take a look elsewhere as well for layout there are a few bits that aren't right still.

shtukas: @adamnfish Corrected! (There were also missing spaces in function argument lists in the tests)

There _is_ a style guide for Scala:  https://docs.scala-lang.org/style/ , I have never read it before....

PR #23, opened 2018-02-15
getRelevantInstance -> getSSHInstance

shtukas:
This change renames `getRelevantInstance` into `getSSHInstance`, modifies its internal logic to handle the case of `Instance`s without an IP address and updates the tests.

### DIFF
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -34,13 +34,22 @@ object Logic {
   }
 
   def getSSHInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
-    val instancesWithPublicIP = instances.filter(_.publicIpAddressOpt.isDefined)
-    if (instancesWithPublicIP.isEmpty) {
+    val sortedValidInstances = instances
+      .filter(_.publicIpAddressOpt.isDefined)
+      .sortBy(_.id.id)
+    if (instances.isEmpty) {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
-      val sortedInstances = instancesWithPublicIP.sortBy(_.id.id)
-      if (instancesWithPublicIP.length==1 || takeAnySingleInstance) Right(sortedInstances.head)
-      else Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedInstances.mkString(", ")}", UnhandledError, None, None)))
+      sortedValidInstances match {
+        case Nil =>
+          Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
+        case instance :: Nil =>
+          Right(instance)
+        case instance :: _ if takeAnySingleInstance =>
+          Right(instance)
+        case _ :: _ =>
+          Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedValidInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
+      }
     }
   }
 
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -11,7 +11,6 @@ import com.gu.ssm.utils.attempt.{ErrorCode, FailedAttempt, Failure, UnhandledErr
 
 import scala.io.Source
 
-
 object Logic {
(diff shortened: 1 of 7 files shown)


shtukas:
Fully tested on local :)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -34,14 +33,13 @@ object Logic {
...
+    val instancesWithPublicIP = instances.filter(_.publicIpAddressOpt.isDefined)
+    if (instancesWithPublicIP.isEmpty) {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
-    } else if (instances.length == 1) {
-      Right(instances.sortBy(_.id.id).head) // head safe as length is necessary > 0
     } else {
-      val sortedInstances = instances.sortBy(_.id.id)
-      if (takeAnySingleInstance) Right(sortedInstances.head) // head safe as length is necessary > 0
+      val sortedInstances = instancesWithPublicIP.sortBy(_.id.id)
+      if (instancesWithPublicIP.length==1 || takeAnySingleInstance) Right(sortedInstances.head)
```
adamnfish: A few things here. I realise they aren't all from this change, but this fn is changing significantly so let's get it right now.

#### Error messages

I think it should check if any instances were found before it does the filter. This is because we will want to provide different failure messages in these two cases ("no matching instances" vs "no instances with public IPs")

#### head call (minor point)

I know the `head` call is safe because you've checked the length in the if statement.

Because it is safe here this is a minor point, but it's better to avoid `head` entirely. Make sit safer for the future and encourages good functional design by not breaking out of Option unsafely (which in turn makes it simpler to understand the code).
It wouldn't be worth it if it made the code any more complex, but I think it's probably simpler to use `headOption`. Option's `toRight` method can be used to provide a failure instance to handle us not being able to find an `Instance`.

adamnfish: I had a think and off the top of my head I'd have written something like this. Don't mean to be dictating implementation and this is a fiddly match (very dependent on the order of the case statements, which is also not perfect).

Look at how this better addresses the above though. In particular, it gives specific error messages for different cases (like no matching instances vs no instances with public IP addresses).

```scala
def getSSHInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
  val sortedValidInstances = instances
    .filter(_.publicIpAddressOpt.isDefined)
    .sortBy(_.id.id)
  if (instances.isEmpty) {
    Left(???) // couldn't find any instances that match
  } else {
    sortedValidInstances match {
      case instance :: _ if takeAnySingleInstance =>
        Right(instance)
      case instance :: Nil =>
        Right(instance)
      case instance :: _ =>
        Left(???) // found multiple matching instances (add any or be specific)
      case Nil =>
        Left(???) // found $instances but none are valid targets (instances need public IP addresses)
    }
  }
}
```

shtukas:
@adamnfish 

With regards to `getSSHInstance` I decided to adopt your `match` idea; it's not that it's _better_, but maybe slightly more functional and scala-ish (which is always good). That having been said, I present the cases in another order than your draft (an order that somehow, given this very particular function, works better for me -- but this is very personal). Comments are welcome.

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -34,13 +34,22 @@ object Logic {
...
+      .sortBy(_.id.id)
+    if (instances.isEmpty) {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
-      val sortedInstances = instancesWithPublicIP.sortBy(_.id.id)
-      if (instancesWithPublicIP.length==1 || takeAnySingleInstance) Right(sortedInstances.head)
-      else Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedInstances.mkString(", ")}", UnhandledError, None, None)))
+      sortedValidInstances match {
+        case Nil =>
+          Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
```
shtukas: I am happy to be criticised on `instances.map(_.id.id).mkString(", ")` :)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -34,13 +34,22 @@ object Logic {
...
+        case instance :: Nil =>
+          Right(instance)
+        case instance :: _ if takeAnySingleInstance =>
+          Right(instance)
+        case _ :: _ =>
```
adamnfish: This matches any non-empty list. If you want it to be correct for the actual case you're testing it would be `_ :: _ :: _` but given that you aren't using either of the matched values (head or tail) and you've already take care of the other cases, this would be better written as `_`.

I agree this order makes more sense.

PR #24, opened 2018-02-17
-s/--selection newest/oldest

shtukas:
In this commit:

1. We update the definition of an instance to `Instance(id: InstanceId, publicIpAddressOpt: Option[String], launchDateTime: Date)`, thereby adding the instance's Launch DateTime.

2. We are moving from `-a/--any <boolean>` to `-s/--selection <selectionMode>`, where selection mode take values in `["any","oldest","newest"]`. This means that we now have one command line option controlling which instance is selected if more than one instances were found in the case of the `ssh` sub command.

3. The README has been updated to remove some obsolete information, be more simple/useful and to be up to date with the `-s/--selection` switch.

### DIFF
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -66,9 +66,9 @@ object ArgumentParser {
       .action((_, c) => c.copy(mode = Some(SsmSsh)))
       .text("Create and upload a temporary ssh key")
       .children(
-        opt[Boolean]('a', "any").optional()
-          .action((flag, args) => args.copy(takeAnySingleInstance = Some(flag)))
-          .text("Indicates whether the command should run on any single instance"),
+        opt[String]('s', "selection").optional()
+          .action((selectionMode, args) => args.copy(singleInstanceSelectionMode = Some(selectionMode)))
+          .text("Indicates whether the command should run on any single instance")
       )
 
     checkConfig { args =>
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -68,7 +68,12 @@ object ArgumentParser {
       .children(
         opt[String]('s', "selection").optional()
           .action((selectionMode, args) => args.copy(singleInstanceSelectionMode = Some(selectionMode)))
-          .text("Indicates whether the command should run on any single instance")
+          .text("""
+                  |    Indicates that in the case of multiple valid instances one
+                  |    should be chosen to proceed and indicates which method to
+                  |    use. With 'any' selects any instance; with 'newest' selects
+                  |    the newest valid instance and with 'oldest' selects the
+                  |    oldest valid instance.""".stripMargin)
       )
 
     checkConfig { args =>
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -33,7 +33,7 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSSHInstance(instances: List[Instance], singleInstanceSelectionModeOpt: Option[String]): Either[FailedAttempt, Instance] = {
+  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
     val validInstances = instances
       .filter(_.publicIpAddressOpt.isDefined)
     if (instances.isEmpty) {
@@ -43,12 +43,11 @@ object Logic {
         case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
         case instance :: Nil => Right(instance)
         case _ :: _ :: _  => {
-          singleInstanceSelectionModeOpt match {
-            case None => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${validInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
-            case Some("any") => Right(validInstances.sortBy(_.id.id).head)             // safe .head
-            case Some("newest") => Right(validInstances.sortBy(_.launchDateTime).last) // safe .last
-            case Some("oldest") => Right(validInstances.sortBy(_.launchDateTime).head) // safe .head
-            case Some(thing) => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"The value of selection is incorrect, given ${thing}, expected: 'any', 'newest' or 'oldest'", UnhandledError, None, None)))
+          sism match {
+            case SismUnspecified => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${validInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
+            case SismAny => Right(validInstances.sortBy(_.id.id).head)             // safe .head
+            case SismNewest => Right(validInstances.sortBy(_.launchDateTime).last) // safe .last
+            case SismOldest => Right(validInstances.sortBy(_.launchDateTime).head) // safe .head
           }
         }
       }
@@ -64,5 +63,14 @@ object Logic {
 
   def computeIncorrectInstances(executionTarget: ExecutionTarget, instanceIds: List[InstanceId]): List[InstanceId] =
     executionTarget.instances.getOrElse(List()).filterNot(instanceIds.toSet)
+
+  implicit def singleInstanceSelectionModeConversion(mode: Option[String]): SingleInstanceSelectionMode = {
+    mode match {
+      case Some("any") => SismAny
+      case Some("newest") => SismNewest
+      case Some("oldest") => SismOldest
+      case _ => SismUnspecified
+    }
+  }
 
 }
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -33,7 +33,7 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSSHInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
+  def getSSHInstance(instances: List[Instance], singleInstanceSelectionModeOpt: Option[String]): Either[FailedAttempt, Instance] = {
     val sortedValidInstances = instances
       .filter(_.publicIpAddressOpt.isDefined)
       .sortBy(_.id.id)
@@ -45,7 +45,7 @@ object Logic {
           Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
         case instance :: Nil =>
           Right(instance)
-        case instance :: _ if takeAnySingleInstance =>
+        case instance :: _ if singleInstanceSelectionModeOpt.getOrElse("") == "any" =>
           Right(instance)
         case _ :: _ :: _ =>
           Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedValidInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -34,21 +34,23 @@ object Logic {
   }
 
   def getSSHInstance(instances: List[Instance], singleInstanceSelectionModeOpt: Option[String]): Either[FailedAttempt, Instance] = {
-    val sortedValidInstances = instances
+    val validInstances = instances
       .filter(_.publicIpAddressOpt.isDefined)
-      .sortBy(_.id.id)
     if (instances.isEmpty) {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
-      sortedValidInstances match {
-        case Nil =>
-          Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
-        case instance :: Nil =>
-          Right(instance)
-        case instance :: _ if singleInstanceSelectionModeOpt.getOrElse("") == "any" =>
-          Right(instance)
-        case _ :: _ :: _ =>
-          Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${sortedValidInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
+      validInstances match {
+        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
+        case instance :: Nil => Right(instance)
+        case _ :: _ :: _  => {
+          singleInstanceSelectionModeOpt match {
+            case None => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${validInstances.map(_.id.id).mkString(", ")}", UnhandledError, None, None)))
+            case Some("any") => Right(validInstances.sortBy(_.id.id).head)             // safe .head
+            case Some("newest") => Right(validInstances.sortBy(_.launchDateTime).last) // safe .last
+            case Some("oldest") => Right(validInstances.sortBy(_.launchDateTime).head) // safe .head
+            case Some(thing) => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"The value of selection is incorrect, given ${thing}, expected: 'any', 'newest' or 'oldest'", UnhandledError, None, None)))
+          }
+        }
       }
     }
   }
--- a/src/main/scala/com/gu/ssm/models.scala
+++ b/src/main/scala/com/gu/ssm/models.scala
@@ -5,10 +5,11 @@ import com.amazonaws.services.ec2.AmazonEC2Async
 import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceAsync
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementAsync
 import com.gu.ssm.aws.SSM
+import java.util.Date
 
 
 case class InstanceId(id: String) extends AnyVal
-case class Instance(id: InstanceId, publicIpAddressOpt: Option[String])
+case class Instance(id: InstanceId, publicIpAddressOpt: Option[String], launchDateTime: Date)
 case class AppStackStage(app: String, stack: String, stage: String)
 case class ExecutionTarget(instances: Option[List[InstanceId]] = None, ass: Option[AppStackStage] = None)
 
--- a/src/main/scala/com/gu/ssm/models.scala
+++ b/src/main/scala/com/gu/ssm/models.scala
@@ -4,7 +4,6 @@ import com.amazonaws.regions.{Region, Regions}
 import com.amazonaws.services.ec2.AmazonEC2Async
 import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceAsync
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementAsync
-import com.gu.ssm.aws.SSM
 import java.util.Date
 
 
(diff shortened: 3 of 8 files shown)


src/main/scala/com/gu/ssm/models.scala
```diff
@@ -4,17 +4,17 @@ import com.amazonaws.regions.{Region, Regions}
...
 
 
 case class InstanceId(id: String) extends AnyVal
-case class Instance(id: InstanceId, publicIpAddressOpt: Option[String])
+case class Instance(id: InstanceId, publicIpAddressOpt: Option[String], launchDateTime: Date)
 case class AppStackStage(app: String, stack: String, stage: String)
 case class ExecutionTarget(instances: Option[List[InstanceId]] = None, ass: Option[AppStackStage] = None)
 
-case class Arguments(executionTarget: Option[ExecutionTarget], toExecute: Option[String], profile: Option[String], region: Region, mode: Option[SsmMode], takeAnySingleInstance: Option[Boolean])
+case class Arguments(executionTarget: Option[ExecutionTarget], toExecute: Option[String], profile: Option[String], region: Region, mode: Option[SsmMode], singleInstanceSelectionMode: Option[String])
```
aug24: Shouldn't this be 

singleInstanceSelectionMode: Option[SingleInstanceSelectionMode]) ?

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -67,7 +68,16 @@ object ArgumentParser {
       .text("Create and upload a temporary ssh key")
       .children(
         opt[String]('s', "selection").optional()
-          .action((selectionMode, args) => args.copy(singleInstanceSelectionMode = Some(selectionMode)))
+            .validate(
+              sism =>
```
shtukas: Nice, I wasn't aware of `.validate`. Yes, pushing `singleInstanceSelectionModeConversion` as early as possible makes perfect sense.

shtukas:
Noice!

shtukas:
I will let Adam have a look before I merge as this was a self imposed admission test to the @adamnfish 's Coding Jedi Academy (^_^)

src/main/scala/com/gu/ssm/ArgumentParser.scala
aug24: +1 for pressing '.' and then reading what is offered :)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -63,4 +64,13 @@ object Logic {
   def computeIncorrectInstances(executionTarget: ExecutionTarget, instanceIds: List[InstanceId]): List[InstanceId] =
     executionTarget.instances.getOrElse(List()).filterNot(instanceIds.toSet)
 
+  implicit def singleInstanceSelectionModeConversion(mode: String): Option[SingleInstanceSelectionMode] = {
```
adamnfish: This should *definitely* not be implicit! It's called explicitly for a start, but even ignoring that `implicit def`s are very dangerous.

This says to the Scala compiler "whenever you need an `Option[SingleInstanceSelectionMode]` and have a `String`, you may call this function to do the conversion for me. This means you could now pass any String into a function expecting an `Option[SingleInstanceSelectionMode]` and as long as this is in scope it'll work. If the value happens to be `"any"`, `"newest"` or `"oldest"` you'll get a Some value, otherwise a None. It breaks type-safety!

`IntelliJ` highlights this and says "Advanced language feature", which is understatement here. To be avoided unless there's a really good reason!

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,22 +33,23 @@ object Logic {
...
+      val validInstances = instances
+        .filter(_.publicIpAddressOpt.isDefined)
+      val validInstanceOrdered = sism match {
+        case Some(SismAny) => validInstances.sortBy(_.id.id)
+        case Some(SismNewest) => validInstances.sortBy(_.launchDateTime).reverse
+        case Some(SismOldest) => validInstances.sortBy(_.launchDateTime)
+        case None => validInstances
+      }
+      validInstanceOrdered match {
+        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
```
adamnfish: It feels like this match to some extent now duplicates the one above. I'm sure this will be simpler if you match on both at once?

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -66,9 +67,23 @@ object ArgumentParser {
...
+                singleInstanceSelectionModeConversion(sism) match {
+                  case None => Left("Unknown instance selection mode: $sism")
+                  case Some(_) => Right(Unit)
+                }
+            )
+          .action((selectionMode, args) => {
+            args.copy(singleInstanceSelectionMode = singleInstanceSelectionModeConversion(selectionMode))
+          })
+          .text("""
+                  |    Indicates that in the case of multiple valid instances one
```
adamnfish: To me this is a sign that this argument is too complicated. Why not just have `--any`, `--newest` and `--oldest`?

Terseness cannot be the argument because in all cases the above is shorter to type than even the condensed `-s` version of what's here in the PR.

p.s. If you agree, I'm not even sure it's worth having the short-form of these. `-n` and especially `-a` and `-o` already have widely used UNIX behaviour that doesn't correspond to our usage. In particular, `-a` usually means `ALL` but means the opposite of that here. (I realise `-a` existed before, but it was weird then as well and if we're overhauling this then now is the time to get it right)

src/main/scala/com/gu/ssm/models.scala
```diff
@@ -48,3 +48,8 @@ case class AWSClients (
 )
 
 case class ResultsWithInstancesNotFound(results: List[(InstanceId, scala.Either[CommandStatus, CommandResult])], instancesNotFound: List[InstanceId])
+
+sealed trait SingleInstanceSelectionMode
```
adamnfish: This would be a better model if it included the `None` case as well rather than was required to be in an `Option`. There are four cases for selection mode, the three here and the default, "fail if more than one is provided". It's not super clear that `None` represents this case, and it makes the types more complicated everywhere. We validate the program at the edge, so let's make sure we have a valid SingleInstanceSelectionMode that capture all the cases at that point.

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
 
   "getRelevantInstance" - {
     import Logic.getSSHInstance
+    import java.text.SimpleDateFormat
+
+    val sip = Some("1278.0.0.1")
```
adamnfish: Not important, but it might be nice to use a valid IP address here?

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
+
+    val formatter = new SimpleDateFormat("yyyy-MM-d HH:mm:ss")
+    val dateOld = formatter.parse("2018-02-17 13:00:14")
```
adamnfish: What's the significance of `old` and `new`?

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
-        }
-
-        "If takeAnySingleInstance is false, should be Left" in {
-          getSSHInstance(twoInstancesWithIP, false).isLeft shouldBe true
+        "should be Left" in {
+          getSSHInstance(twoInstancesWithoutIP, None).isLeft shouldBe true
         }
       }
 
       "At least one instance is well formed" - {
```
adamnfish: This is really saying "exactly one instance is well formed", I think

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
 
   "getRelevantInstance" - {
     import Logic.getSSHInstance
+    import java.text.SimpleDateFormat
+
+    val sip = Some("1278.0.0.1")
+
+    val formatter = new SimpleDateFormat("yyyy-MM-d HH:mm:ss")
+    val dateOld = formatter.parse("2018-02-17 13:00:14")
+    val dateNew = formatter.parse("2018-02-17 15:05:22")
```
adamnfish: Might be helpful to use relative dates here, to show that there's no significance to the date beyond their difference. You could use something like `DateTime.now()` vs `DateTime.now().minusDays(7)` for new and old? It would also be a bit simopler because you'd be able to drop the formatter and datestrings.

src/test/scala/com/gu/ssm/LogicTest.scala
adamnfish: Ah I see, it's used to select the right instance lower down. It might be helpful to set the dates down there with the date tests when you need them. This is irrelevant info for all the other tests but it adds to the complexity up here for everything. You could create the instances closer to where they're needed, make a copy of the data that manipulates it for the specific test, or ignore me and leave it like this. (all fine)

I've mentioned before that it's nice to use FreeSpec's scoping to put test data next the description of why the test data is the way it is. You get the clarity of the groupings, as well as the scoping to setup shared test data. You also avoid the scenario of "create one giant magic test fixture at the top and then test against that", which is the opposite of the more functional "use scalacheck to generate test data and verify properties" or more formal "prove the properties". We don't do much of the latter two, but it's nice to avoid the former. Some problems with it are:

* it couples the tests, which makes it harder to change one of them without affecting others
* so the test data becomes brittle because it's hard to change
* which encourages devs to test specific properties that match the test data, rather than the properties that matetr to the functionality

It's a minor point here :-)

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
+          getSSHInstance(twoInstancesWithoutIP, None).isLeft shouldBe true
         }
       }
 
       "At least one instance is well formed" - {
         val twoMixedInstances = List(instanceYWithoutIP, instanceXWithIP)
 
-        "If takeAnySingleInstance is true, selects the well-formed instance" in {
-          getSSHInstance(twoMixedInstances, true).right.get shouldEqual instanceXWithIP
+        "If single instance selection mode is SismUnspecified), should be Left" in {
```
adamnfish: Your test name here reinforces my point about the choice of model for this feature. `SismUnspecified` would be much better than `None` :-)

adamnfish: What's the `)` for?

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -43,79 +43,100 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
       "All instances are well formed" - {
         val twoInstancesWithIP = List(instanceYWithIP, instanceXWithIP)
 
-        "If takeAnySingleInstance is true, selects the first well-formed instance (in lexicographic order of InstanceId)" in {
-          getSSHInstance(twoInstancesWithIP, true).right.get shouldEqual instanceXWithIP
+        "If single instance selection mode is SismUnspecified, should be Left" in {
+          getSSHInstance(twoInstancesWithIP, None).isLeft shouldBe true
+        }
+
+        "If single instance selection mode is SismAny, selects the first well-formed instance (in lexicographic order of InstanceId)" in {
```
adamnfish: I wonder whether, in the "have instance start time" world, whether alphabetical order would be better as a date order?

(Basically, I am wondering about aliasing `--any` to `--newest`)

src/test/scala/com/gu/ssm/SSHTest.scala
```diff
@@ -50,10 +50,11 @@ class SSHTest extends FreeSpec with Matchers with EitherValues {
 
   "create ssh command" - {
     import SSH.sshCmd
+    import java.util.Date
 
     "create ssh command" in {
       val file: File = new File("/banana")
-      val instance: Instance = Instance(InstanceId("raspberry"), Some("127.0.0.1"))
+      val instance: Instance = Instance(InstanceId("raspberry"), Some("127.0.0.1"), new Date())
```
adamnfish: I don't think we need the type annotation here (there's enough duplication already)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,22 +33,23 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSSHInstance(instances: List[Instance], takeAnySingleInstance: Boolean): Either[FailedAttempt, Instance] = {
-    val sortedValidInstances = instances
-      .filter(_.publicIpAddressOpt.isDefined)
-      .sortBy(_.id.id)
+  def getSSHInstance(instances: List[Instance], sism: Option[SingleInstanceSelectionMode]): Either[FailedAttempt, Instance] = {
```
adamnfish: This function has got very complicated (see below for one suggestion about how to simplify it a little)

src/main/scala/com/gu/ssm/ArgumentParser.scala
shtukas: I thought about this and I strongly feel against it keeping `--any` and adding the others. 

Keeping all three would be the natural thing to do given the pre-existing `--any` but then the user interface designer (eg: me) would have needed to make a completely arbitrary choice of deciding which takes priority in say

```
./ssm ssh -p <profile> -i <instance> --any --newest
```

Now, about terseness, I *totally* agree that sometimes like verbosity and `--selection` takes keystrokes. I am totally up for renaming "selection" into something shorter and moving the values "any", "newest" and "oldest" into shorter things. But giving up the logic of one switch and three values for the sake of aggressive terseness, doesn't work for me.  

Also, remember that `--any`'s true for was `--any true` and `--any false`, still two particles.

ps: It is true that we should come to an agreement because this change will hit the release page :)

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: Sorry that was a typo :)

src/main/scala/com/gu/ssm/Logic.scala
shtukas: TIL You don't like implicits and I have to say that I agree with you! Like totally. I thought I would demonstrate to myself that I know how to use them, but I am happy with getting rid of it. The conversion function is called only once anyway....

shtukas: Little update on this. As I have just noticed, when @aug24 did his amendment this morning, he called the function explicitly, so the implicit wasn't no longer needed. The implicit was introduced by me in the original code when the conversion was (implicitly) done elsewhere.

src/main/scala/com/gu/ssm/models.scala
shtukas: > This would be a better model if it included the None case as well rather than was required to be in an Option.

This actually was the way it was originally, with the `None` case corresponding to an additional object of the sealed trait called `SismUnspecified` but @aug24 convinced me to move from 


```
        case SismAny => validInstances.sortBy(_.id.id)
        case SismNewest => validInstances.sortBy(_.launchDateTime).reverse
        case SismOldest => validInstances.sortBy(_.launchDateTime)
        case SismUnspecified => validInstances
```


```
        case Some(SismAny) => validInstances.sortBy(_.id.id)
        case Some(SismNewest) => validInstances.sortBy(_.launchDateTime).reverse
        case Some(SismOldest) => validInstances.sortBy(_.launchDateTime)
        case None => validInstances
```

The resulting side effect is that signatures that used to have `SingleInstanceSelectionMode` now have `sism: Option[SingleInstanceSelectionMode]`. 

I have to say that both are admittedly good, but I have, in the end, a slight preference for @aug24 's style, even though I totally get where you are coming from and your argument that "We validate the program at the edge" is indeed a *very* good point.

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: That one is funny!

The test names where chosen by me accordingly to the old style, but then the test were updated by @aug24's later on and the test names left as they were (^_^)

> What's the ) for?

Typo. Fixed.

src/test/scala/com/gu/ssm/SSHTest.scala
shtukas: This was there for a long time... Fixed!

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: Good suggestion. Done!

src/main/scala/com/gu/ssm/ArgumentParser.scala
adamnfish: `--any --newest` is an invalid input and would be validated as such!

src/main/scala/com/gu/ssm/models.scala
shtukas: I eventually decided that "We validate the program at the edge" was more important than "slight preference for @aug24 's style". So I essentially reverted @aug24 's changes :)

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: Done, but I didn't find an easy way to shift a Java `Date`, so it ended up being 

```
val dateOld = Date.from(new LocalDateTime().minusDays(7).atZone(ZoneId.systemDefault()).toInstant())
val dateNew = new Date
```

src/main/scala/com/gu/ssm/ArgumentParser.scala
shtukas: > --any --newest is an invalid input and would be validated as such

Thank for pointing that out. I never thought of it like that. Let me sleep on this :)

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: Not really. It appear to the reader that exactly one instance is well formed, but what I mean is "the set of instances is heterogeneous". In other words that there is at least one well formed and at least one ill formed. 

Now, because I am lazy and only defined two instances in my set, the only way to have the above true is one well formed and one ill formed :)

The test progression logic is
- "All instances are ill-formed"
- "At least one instance is well formed" (which should be "Heterogenous set of instances")
- "All instances are well formed"

src/main/scala/com/gu/ssm/Logic.scala
shtukas: If you want to simplify it, you need to simultaneously treat `sism` and the length of `validInstanceOrdered` together, it was like that at first but looked horrible. 

The best I can do, to remain faithful to the current logic and male you happy is to move from
 
```scala
  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
    if (instances.isEmpty) {
    } else {
      val validInstances = instances
        .filter(_.publicIpAddressOpt.isDefined)
      val validInstanceOrdered = sism match {
        case SismNewest => validInstances.sortBy(_.launchDateTime).reverse
        case SismOldest => validInstances.sortBy(_.launchDateTime)
        case SismUnspecified => validInstances
      }
      validInstanceOrdered match {
        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
        case instance :: Nil => Right(instance)
        case _ :: _ :: _ if sism == SismUnspecified => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${validInstances.map(_.id.id).mkString(", ")}.  Use --selection [any|oldest|newest]?", UnhandledError, None, None)))
        case instance :: _ :: _ => Right(instance)
      }
    }
  }
```

to

```scala
  def applyOrderToInstances(instance: List[Instances], sism: SingleInstanceSelectionMode): List[Instance] = (blah blah)

  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
    if (instances.isEmpty) {
    } else {
      val validInstances = instances
        .filter(_.publicIpAddressOpt.isDefined)
      val validInstanceOrdered = applyOrderToInstances(validInstances, sism)
      validInstanceOrdered match {
        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
        case instance :: Nil => Right(instance)
        case _ :: _ :: _ if sism == SismUnspecified => Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Error choosing single instance, found ${validInstances.map(_.id.id).mkString(", ")}.  Use --selection [any|oldest|newest]?", UnhandledError, None, None)))
        case instance :: _ :: _ => Right(instance)
      }
    }
  }
```

src/test/scala/com/gu/ssm/LogicTest.scala
shtukas: > I've mentioned before that it's nice to use FreeSpec's scoping to put test data next the description of why the test data is the way it is.

@adamnfish I have just made an update you will like. You see, you never mentioned I could -_cheat_ :)

In the test suite I defined 

```scala
def makeInstance(id: String, ipOpt:Option[String], launchDateDayShift: Int) =
      Instance(InstanceId(id), ipOpt, Date.from(LocalDateTime.now().plusDays(launchDateDayShift).atZone(ZoneId.systemDefault()).toInstant()))
```

With this done, the suite is much more like you wanted: readable and "localised". Example

```
      "At least one instance is well formed" - {
        val i1 = makeInstance("X", None, -7)
        val i2 = makeInstance("Y", Some("127.0.0.1"), -1)
        val i3 = makeInstance("Z", Some("127.0.0.1"), 0)

        "If single instance selection mode is SismNewest, selects the newest well-formed instance" in {
          getSSHInstance(List(i1, i2, i3), SismNewest).right.get shouldEqual i3
        }

        "If single instance selection mode is SismOldest, selects the oldest well-formed instance" in {
          getSSHInstance(List(i1, i2, i3), SismOldest).right.get shouldEqual i2
        }

        "If single instance selection mode is SismUnspecified, should be Left" in {
          getSSHInstance(List(i1, i2, i3), SismUnspecified).isLeft shouldBe true
        }
      }
```

PR #28, opened 2018-02-20
Better introduction to what ssm-scala is

shtukas:


README.md
```diff
@@ -1,10 +1,12 @@
...
 =========
 
-SSM-Scala is a command-line tool for executing commands on EC2 servers
-using EC2 Run Command, written in Scala. This can be used as an
-alternative to `ssh` for servers in AWS accounts to which you have
-[IAM](https://aws.amazon.com/iam/) access.
+SSM-Scala is a command-line tool, written in Scala, for executing commands on EC2 servers using AWS's EC2 Run command. It provides the user with 
+
+- standard `ssh` access using short lived RSA keys, and
+- an _alternative_ to `ssh` for running commands on the target.
```
katebee: Nit: The combination of bullet points and sentence structure feels odd, but maybe that is just me? ¯\\\_(ツ)\_/¯ 

I would go for:

> SSM-Scala is a command-line tool, written in Scala, for executing commands on EC2 servers using AWS's EC2 Run command. It provides the user with:

> 1. standard ssh access using short lived RSA keys
> 2. an alternative to ssh for running commands on the target

> Both modes apply to servers in AWS accounts to which you have IAM access.

Or alternatively:

> SSM-Scala is a command-line tool, written in Scala, for executing commands on EC2 servers using AWS's EC2 Run command. It provides the user with standard ssh access using short lived RSA keys, and an alternative to ssh for running commands on the target.

> Both modes apply to servers in AWS accounts to which you have IAM access.

Opinion not requirement! - YMMV 😁

shtukas:
@katebee 

But I will go for your first proposal, because of more familiarity with that style :)

PR #30, opened 2018-02-20
Use private IP address if no public IP address

### LINKED ISSUE #29: Add support for private IP addresses
Most of our instances do not have public IP addresses and we connect directly to them on their private IP addresses via a VPN gateway from the corporate internal network. As such`ssm` should be able to generate the connection string for instances that only have private IP addresses.

satterly:
Fixes #29


src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,7 +45,9 @@ object Main {
       instance <- Attempt.fromEither(Logic.getSSHInstance(config.targets, sism))
       _ <- IO.tagAsTainted(instance.id, config.name, awsClients.ec2Client)
       _ <- IO.installSshKey(instance.id, config.name, addAndRemoveKeyCommand, awsClients.ssmClient)
-    } yield SSH.sshCmd(authFile, instance)
+      publicOrPrivateIp = (instance.publicIpAddressOpt ++ instance.privateIpAddressOpt).headOption
```
adamnfish: This would be better written with `orElse`, I think?

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -38,7 +38,7 @@ object Logic {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
       val validInstancesWithOrder = instances
-        .filter(_.publicIpAddressOpt.isDefined)
+        .filter(x => x.publicIpAddressOpt.isDefined || x.privateIpAddressOpt.isDefined)
```
adamnfish: We could probably come up with a better name than `x` (`i`, for example?)

src/main/scala/com/gu/ssm/models.scala
```diff
@@ -7,13 +7,33 @@ import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
...
+  executionTarget: Option[ExecutionTarget],
+  toExecute: Option[String],
+  profile: Option[String],
+  region: Region,
+  mode: Option[SsmMode],
+  singleInstanceSelectionMode: SingleInstanceSelectionMode,
+  isSelectionModeNewest: Boolean,
+  isSelectionModeOldest: Boolean,
+  usePrivateIpAddress: Boolean
+)
```
satterly: Happy to undo this whitespace change if wanted. This line and the one below were getting pretty long though.

src/main/scala/com/gu/ssm/Main.scala
satterly: Probably you're right. It's been changed to a pattern match now though.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,13 +45,21 @@ object Main {
...
-    } yield SSH.sshCmd(authFile, instance)
+      ipAddress <- Attempt.fromEither(getIpAddress(instance, usePrivate))
+    } yield SSH.sshCmd(authFile, instance, ipAddress)
 
     val programResult = Await.result(fProgramResult.asFuture, maximumWaitTime)
     programResult.fold(UI.outputFailure, UI.sshOutput)
     System.exit(programResult.fold(_.exitCode, _ => 0))
   }
 
+  private def getIpAddress(instance: Instance, usePrivate: Boolean): Either[FailedAttempt, String] = (instance.publicIpAddressOpt, instance.privateIpAddressOpt) match {
```
adamnfish: Main is for the CLI program and the sub-programs it spawns. This is just a piece of pure logic so it'd be better placed elsewhere. I'd recommend the `Logic` object since this is just logic.

I'd alsohope to see tests for this function because it is both important and fiddly.

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,13 +45,21 @@ object Main {
...
+    case (Some(pub), _) if !usePrivate => Right(pub)
```
adamnfish: It feels like these cases are duplicating each other. Would it be more simply written as something like this?

```scala
if (usePrivate) {
  instance.privateIpAddressOpt.toRight( .. private IP failure .. )
} else {
  instance.publicIpAddressOpt.toRight( .. public IP failure .. )
}
```

src/main/scala/com/gu/ssm/models.scala
adamnfish: Yep agreed, definitely large enough to warrant this change. Thanks!

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -80,6 +80,12 @@ object ArgumentParser {
               isSelectionModeOldest = true)
           })
           .text("Selects the oldest instance if more than one instance was specified"),
+        opt[Unit]('p', "private").optional()
+          .action((_, args) => {
+            args.copy(
+              usePrivateIpAddress = true)
+          })
+          .text("Use private IP address"),
```
adamnfish: Could this be a teeny bit more informative? (Nice that it is very terse)

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -80,6 +80,12 @@ object ArgumentParser {
               isSelectionModeOldest = true)
           })
           .text("Selects the oldest instance if more than one instance was specified"),
+        opt[Unit]('p', "private").optional()
```
adamnfish: `p` clashes with `--profile`. Does this definitely need a short form? (to be fair, I don't think profile necessarily needs a short form either)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -38,7 +38,7 @@ object Logic {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
       val validInstancesWithOrder = instances
-        .filter(_.publicIpAddressOpt.isDefined)
+        .filter(i => i.publicIpAddressOpt.isDefined || i.privateIpAddressOpt.isDefined)
```
adamnfish: I think this filter is now pointless because all instances have a private IP address, and this highlights that the change breaks this part of the logic.

This function's job is to pick a "valid" instance. Previously that meant selecting an instance with a public IP address. The measure of "valid" now depends on whether the private flag was set so this function probably needs to take that flag and it should return the first instance that matches the dynamic requirement.

The now-broken behaviour is this:
At the moment this function will select an instance and then you'll apply another check to look for the correct (public or private) IP address on what gets returned. It may be that the instance selected here does not meet the requirements, but there was an instance among those that were matched that could have. With this change we now erroneously fail.

src/main/scala/com/gu/ssm/ArgumentParser.scala
satterly: Slightly surprised the argument parser didn't complain about this.

src/main/scala/com/gu/ssm/Main.scala
satterly: I assume I don't need a test for the pattern match now as it's a simple `if` statement (see below).

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,12 +33,12 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
+  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode, usePrivate: Boolean): Either[FailedAttempt, Instance] = {
```
adamnfish: Be good to see some tests for this changed behaviour.

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -60,4 +60,12 @@ object Logic {
   def computeIncorrectInstances(executionTarget: ExecutionTarget, instanceIds: List[InstanceId]): List[InstanceId] =
     executionTarget.instances.getOrElse(List()).filterNot(instanceIds.toSet)
 
+  def getIpAddress(instance: Instance, usePrivate: Boolean): Either[FailedAttempt, String] = {
```
adamnfish: Still be nice to have some tests for this function.

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -60,4 +60,12 @@ object Logic {
...
+    if (usePrivate) {
+      instance.privateIpAddressOpt.toRight(FailedAttempt(Failure("No private IP address", "No private IP address and '--private' option used", NoIpAddress, None, None)))
+    } else {
+      instance.publicIpAddressOpt.toRight(FailedAttempt(Failure("No public IP address", "No public IP address", NoIpAddress, None, None)))
```
adamnfish: I like the message above, could this one also refer to the `private` flag? Maybe Something along the lines of:

    "No public IP address (use `--private` to specify private IP)"

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,15 +33,15 @@ object Logic {
...
   }
 
-  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
+  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode, usePrivate: Boolean = false): Either[FailedAttempt, Instance] = {
     if (instances.isEmpty) {
       Left(FailedAttempt(Failure(s"Unable to identify a single instance", s"Could not find any instance", UnhandledError, None, None)))
     } else {
       val validInstancesWithOrder = instances
-        .filter(_.publicIpAddressOpt.isDefined)
+        .filter(i => usePrivate || (!usePrivate && i.publicIpAddressOpt.isDefined))
```
adamnfish: isn't this the same as the following? (you're testing usePrivate twice):

    .filter(i => usePrivate || i.publicIpAddressOpt.isDefined)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,15 +33,15 @@ object Logic {
...
         .sortBy(_.launchInstant)
       validInstancesWithOrder match {
-        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets (instances need public IP addresses)", UnhandledError, None, None)))
+        case Nil => Left(FailedAttempt(Failure(s"Instances with no IPs", s"Found ${instances.map(_.id.id).mkString(", ")} but none are valid targets", UnhandledError, None, None)))
```
adamnfish: `"valid targets"` <- I'm not 100% sure this will make sense to a user. We added the note about IP addresses to make this clearer, shame to lose it now.

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -44,8 +44,8 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
   "getRelevantInstance" - {
```
adamnfish: `getRelevantInstance` now takes a new parameter, so we'll need to test that new behaviour. With the "always has a private IP" change there's not heaps that needs testing, but need to show that he argument is important.

(at the moment you could replace the argument with `true` /`false` and all the tests would still pass)

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -33,15 +33,15 @@ object Logic {
     case _ => Right(Unit)
   }
 
-  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance] = {
+  def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode, usePrivate: Boolean = false): Either[FailedAttempt, Instance] = {
```
adamnfish: usePrivate

adamnfish:
Also spotted the function has changed name from `getRelevantInstance` to `getSSHInstance`. Can we update the test accordingly?

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -91,15 +91,67 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
         "If single instance selection mode is SismUnspecified, should be Left" in {
-          getSSHInstance(List(i1, i2, i3), SismUnspecified).isLeft shouldBe true
+          getSSHInstance(List(i1, i2, i3), SismUnspecified, usePrivate = false).isLeft shouldBe true
+        }
+      }
+    }
+
+    "Given one instance and want private IP" - {
+
+      "Instance is ill-formed should be Left" in {
```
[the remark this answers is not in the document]
adamnfish: This is passing because you give it an empty list (which contradicts the message). Better to remove this case since it doesn't make sense here.

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -91,15 +91,67 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
...
+        "If single instance selection mode is SismUnspecified, returns argument" in {
+          val i = makeInstance("X", Some("127.0.0.1"), "10.1.1.10", 0)
+          getSSHInstance(List(i),  SismUnspecified, usePrivate = true).right.get shouldEqual i
+        }
+      }
+    }
+
+    "Given more than one instance and want private IP" - {
+
+      "All instances are ill-formed, should be Left" in {
```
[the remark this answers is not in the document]
adamnfish: This fails because there are multiple instances provided, not because they are ill formed. That case is tested below, so better to remove this case (since it doesn't make sense here).

PR #32, opened 2018-02-22
Adding LICENSE and NOTICE files

shtukas:


NOTICE
```diff
@@ -0,0 +1,14 @@
+   SSM-SCALA
+   Copyright 2011-2018 Guardian News & Media Ltd
```
satterly: This should be `2018` only, not `2011-2018`

satterly:
It's good practice to make reference to the license in the README but this isn't essential. It just saves people a bit time hunting for it. Something like this at the bottom of README files is common ...

## License

Copyright (c) 2018 Guardian News & Media. Available under the Apache License.

shtukas:
Thanks @satterly I implemented both suggestions.

PR #33, opened 2018-02-23
Poc script for reporting

aug24:


shtukas:
@aug24 You probably already know that, but just for the record:

```
$ ./ssh-report security -1H

An error occurred (ExpiredTokenException) when calling the ListCommands operation: The security token included in the request is expired
```

shtukas:
As discussed, not very _useful_ but a perfect example of what can be done with the aws command line and the shell script is awesome. Independent from the main code and might evolve later. +1 from me

shtukas:
Yeepee!

PR #34, opened 2018-02-26
Introduce InvocationDoesNotExist and handle AWS error

amyhughes:
We were finding the following error when running a command against all instances of our elasticsearch cluster:

```
$  ~ ./ssm cmd -c 'date' --profile ophan -t elasticsearch,ophan,PROD
SLF4J: A number (14) of logging calls during the initialization phase have been intercepted and are
SLF4J: now being replayed. These are subject to the filtering rules of the underlying logging system.
SLF4J: See also http://www.slf4j.org/codes.html#replay
Unknown error while making an API call to AWS' AWSSimpleSystemsManagement service, null (Service: AWSSimpleSystemsManagement; Status Code: 400; Error Code: InvocationDoesNotExist; Request ID: d46022a8-188f-11e8-9548-511ec617bfed)
```
We fix this by capturing this particular error from AWS as a new `CommandStatus`: `InvocationDoesNotExist`.

We have tested this change locally and can confirm we no longer receive the error. 
@rtyley

### DIFF
--- a/src/main/scala/com/gu/ssm/aws/SSM.scala
+++ b/src/main/scala/com/gu/ssm/aws/SSM.scala
@@ -39,7 +39,11 @@ object SSM {
     val request = new GetCommandInvocationRequest()
       .withCommandId(commandId)
       .withInstanceId(instance.id)
-    handleAWSErrs(awsToScala(client.getCommandInvocationAsync)(request).map(extractCommandResult))
+    handleAWSErrs(
+      awsToScala(client.getCommandInvocationAsync)(request)
+        .map(extractCommandResult)
+        .recover { case e if e.getMessage.contains("InvocationDoesNotExist") => Left(InvocationDoesNotExist) }
+    )
   }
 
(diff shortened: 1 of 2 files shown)


shtukas:
Absolutely fantastic. Thanks @amyhughes :)

shtukas:
Fully tested on local. Merging now.

src/main/scala/com/gu/ssm/aws/SSM.scala
```diff
@@ -39,7 +39,11 @@ object SSM {
     val request = new GetCommandInvocationRequest()
       .withCommandId(commandId)
       .withInstanceId(instance.id)
-    handleAWSErrs(awsToScala(client.getCommandInvocationAsync)(request).map(extractCommandResult))
+    handleAWSErrs(
+      awsToScala(client.getCommandInvocationAsync)(request)
+        .map(extractCommandResult)
+        .recover { case e if e.getMessage.contains("InvocationDoesNotExist") => Left(InvocationDoesNotExist) }
```
rtyley: You can detect this by type rather than message text - see https://github.com/guardian/ssm-scala/pull/36 !

PR #35, opened 2018-02-26
[PoC] Add `emr-cluster` flag to support running commands on EMR master nodes

desbo:
This adds a new flag to facilitate executing commands on EMR master instances. This will be useful for e.g. using the interactive Hive shell on an instance.  I've also done a touch of refactoring (mainly to split out `ExecutionTarget` to  a trait instead of a single class with optional fields).

Note that I haven't tested anything yet, this is just to get some feedback on the basic idea.


shtukas:
@desbo 
Sorry for the delay in approving this. I prefer @adamnfish to do it but he's been away (back tomorrow). Just pointing out to you that your PR is conflicting with master, could you have a look ? Cheers,

desbo:
Thanks @shtukas – after thinking about it some more, we're going to address this issue in a different way that won't require any changes to SSM, so I'm closing this.

shtukas:
Ok cool.

PR #36, opened 2018-02-27
Detect `InvocationDoesNotExistException` by type rather than message text

rtyley:
We can take advantage of the strongly-typed `InvocationDoesNotExistException` provided by the AWS SDK. This doesn't make the fix introduced with d9d691e in #34 any better, just a bit more concise, and more strongly-typed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,9 +109,9 @@ Note that if the argument `-t <app>,<stack>,<stage>` resolves to more than one i
 During development, the program can be run using sbt, either from an
 sbt shell or from the CLI in that project.
 
-    $ sbt "run --c pwd --instances i-0123456 --profile xxx --region xxx"
+    $ sbt "run cmd -c pwd --instances i-0123456 --profile xxx --region xxx"
 
-    sbt:ssm-scala> run --c pwd --instances i-0123456 --profile xxx --region xxx 
+    sbt:ssm-scala> run cmd -c pwd --instances i-0123456 --profile xxx --region xxx 
 
 However, `sbt` traps the program exit so in REPL mode you may find it
 easier to create and run an executable instead, for this just run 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -109,9 +109,9 @@ Note that if the argument `-t <app>,<stack>,<stage>` resolves to more than one i
 During development, the program can be run using sbt, either from an
 sbt shell or from the CLI in that project.
 
-    $ sbt "run --c pwd --instances i-0123456 --profile xxx --region xxx"
+    $ sbt "run cmd -c pwd --instances i-0123456 --profile xxx --region xxx"
```
shtukas: Many thanks! And for the other one as well :)

src/main/scala/com/gu/ssm/aws/SSM.scala
```diff
@@ -42,7 +42,7 @@ object SSM {
     handleAWSErrs(
       awsToScala(client.getCommandInvocationAsync)(request)
         .map(extractCommandResult)
-        .recover { case e if e.getMessage.contains("InvocationDoesNotExist") => Left(InvocationDoesNotExist) }
+        .recover { case _:InvocationDoesNotExistException => Left(InvocationDoesNotExist) }
```
shtukas: Perfect. I wasn't aware of that. Very nice!

shtukas:
Works fine on my local. I could not test the change against a large number of instances because I lost my Ophan credentials yesterday, but I have no doubt that you did. 

Merge at will.

rtyley:
> Works fine on my local. I could not test the change against a large number of instances because I lost my Ophan credentials yesterday, but I have no doubt that you did.

Thanks man - yeah, you're right, I did check, but I should have mentioned that! I used this expression to check it worked:

```
$ sbt "run cmd -c pwd -t elasticsearch,ophan,PROD --profile ophan --region eu-west-1"`
```

PR #37, opened 2018-02-27
New installation procedures and developer release notes

shtukas:
Update the README with a new installation procedure using `brew` and two tasks to sync the latest version with the Guardian tap.

Thanks @sihil for the idea and work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,13 @@ Instructions for using SSM Scala in your own project can be found [below](#How-t
 
 ## Installation
 
+If you have Homebrew installed do
+
 ```
 brew install guardian/devtools/ssm
 ```
+
+Otherwise, fetch the most recently released version of the program from the [Github releases page](https://github.com/guardian/ssm-scala/releases/latest) and make sure it is executable (`chmod +x ssm`). You may then want to put it somewhere in your PATH.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,9 @@ Instructions for using SSM Scala in your own project can be found [below](#How-t
 
 ## Installation
 
-Fetch the most recently released version of the program from the [Github releases page](https://github.com/guardian/ssm-scala/releases/latest) and make sure it is executable (`chmod +x ssm`). You may then want to put it somewhere in your PATH.
```
katebee: We might want to leave this in? Could it be helpful to anyone on anything not Mac OS?

README.md
```diff
@@ -12,7 +12,9 @@ Instructions for using SSM Scala in your own project can be found [below](#How-t
...
+```
```
adamnfish: This only applies to OSX users that use brew. I agree adding this is great and I also agree it meets our primary use-case, but I think we still need something to explain the old way for everyone else (even if that is a simplified description of what was there).

adamnfish: MVP:
```
For non-brew users you can fetch [the most recently release from the Github releases page](https://github.com/guardian/ssm-scala/releases/latest).
```

shtukas:
@katebee @adamnfish Thanks for pointing that out. Corrected. Cheers,

PR #38, opened 2018-02-28
Add support for specifying ssh user

satterly:
**Example**

```
$ ssm ssh -i i-066ea79153a164de8 --profile ophan --private --user hadoop
```

### DIFF
--- a/src/main/scala/com/gu/ssm/models.scala
+++ b/src/main/scala/com/gu/ssm/models.scala
@@ -17,18 +17,22 @@ case class Arguments(
   profile: Option[String],
   region: Region,
   mode: Option[SsmMode],
+  user: Option[String],
   singleInstanceSelectionMode: SingleInstanceSelectionMode,
   isSelectionModeNewest: Boolean,
   isSelectionModeOldest: Boolean,
   usePrivateIpAddress: Boolean
 )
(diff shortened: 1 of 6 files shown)


shtukas:
@satterly 
Do you mind if I replace `-U, --user <user>` with `-u, --user <user>` ?

shtukas:
Awesome. Thanks!

src/main/scala/com/gu/ssm/models.scala
```diff
@@ -17,18 +17,22 @@ case class Arguments(
...
   region: Region,
   mode: Option[SsmMode],
+  user: Option[String],
   singleInstanceSelectionMode: SingleInstanceSelectionMode,
   isSelectionModeNewest: Boolean,
   isSelectionModeOldest: Boolean,
   usePrivateIpAddress: Boolean
 )
 object Arguments {
+  val DefaultUser = "ubuntu"
```
adamnfish: Should be lower case for a val `defaultUser`

shtukas: True!

shtukas: https://github.com/guardian/ssm-scala/pull/45

PR #40, opened 2018-03-06
Respect AWS_PROFILE environment variable

aug24:


shtukas:
I didn't test it, but looks good. Also you might have just fulfilled an outstanding user request :)

aug24:
I hadn't noticed Simon's issue, but it was annoying me too!

We still don't respect AWS_DEFAULT_PROFILE, so perhaps we should add code to check for that too.

shtukas:
Maybe they use AWS_DEFAULT_PROFILE for some other internal tools. As far as ssm in concerned let's stick to AWS_PROFILE.

PR #42, opened 2018-03-06
Introduces --raw

shtukas:
This change introduces the `--raw` flag for the ssh subcommand allowing the output to be used as part of a pipe. For instance

```
ssm ssh --profile security -t security-hq,security,PROD --newest --raw | bash
```

will execute the ssh connection directly. 

Note that we moved from 

```
sshOutput(result: (InstanceId, String))
``` 

to

```
sshOutput(machineOutput: Boolean)(result: (InstanceId, String))
``` 

as the best way to keep `setUpSSH`'s `programResult.fold` clean.

PR #43, opened 2018-03-07
Handle pseudo-terminal for raw

aug24:
## What does this change?

Forces pseudo terminal allocation for --raw invocation

## What is the value of this?

--raw is usually piped to bash, which requires a pseudo terminal to be forced

## Any additional notes?

No


shtukas:
Perfect! Thanks @aug24

shtukas:
Follow up change after: https://github.com/guardian/ssm-scala/pull/42#issuecomment-370797601

PR #45, opened 2018-03-07
Correct variable capitalization

shtukas:


shtukas:
@adamnfish

satterly:
Not according to scala style guide. https://docs.scala-lang.org/style/naming-conventions.html

shtukas:
Um........ That's an interesting one... 
I leave this open further notice (^_^)

adamnfish:
Fair play, that says use a different convention for constants and this could qualify as such on account of it belonging to an object.

The reason not to do that is that Scala treats capital names as types and little names as values. It means that, for example, you can't put this value in a match statement.

The rule of "types are capitals and values are lower case" conflicts with the ruling in the style guide. Since the style guide case you cite is an edge case for "constants", and given that there are concrete reasons in the language not to name values with upper case I would follow the rule of types / values. It's consistent with the rest of the codebase and means a reader (especially outside of the IDE) can be confident a capital name is a type.

shtukas:
I agree with @adamnfish, that was my first mental reaction when I saw his original comment.

I self :+1: that one. Will merge later unless a better argument comes up

PR #46, opened 2018-03-07
Remove bashisms

aug24:
## What does this change?

Posix compliance - does not rely on root having a bash shell by default

## What is the value of this?

Will work on machines which do not use bash for root.

## Any additional notes?

Also introduces a one second delay before output in "--raw | bash" mode, which should prevent the race condition between adding the key and invoking the command in bash.

### DIFF
--- a/src/main/scala/com/gu/ssm/UI.scala
+++ b/src/main/scala/com/gu/ssm/UI.scala
@@ -24,6 +24,7 @@ object UI {
 
   def sshOutput(rawOutput: Boolean)(result: (InstanceId, String)): Unit = {
     if (rawOutput){
+      Thread.sleep(1000)
       print(result._2)
     } else {
       UI.printMetadata(s"========= ${result._1.id} =========")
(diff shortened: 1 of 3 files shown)


src/main/scala/com/gu/ssm/UI.scala
```diff
@@ -24,6 +24,7 @@ object UI {
 
   def sshOutput(rawOutput: Boolean)(result: (InstanceId, String)): Unit = {
     if (rawOutput){
+      Thread.sleep(1000)
```
shtukas: I don't feel super comfortable with such hard coed value as it may one day not be enough by few tenths of a second and break somebody's script, but if that's what needed....

shtukas:
+1 here.

src/main/scala/com/gu/ssm/UI.scala
aug24: I've got nothing better.  As discussed, we can't wait for the command to finish, because the command deliberately incorporates the removal as well as the insertion of the temporary key.

shtukas: +1

PR #47, opened 2018-03-07
Support domain names

aug24:
## What does this change?

Preferentially uses public dns if the instance has one

## What is the value of this?

Integrates better with users' ssh configurations.

## Any additional notes?

No


shtukas:
Nice one :)

shtukas:
+1

PR #51, opened 2018-04-03
Bastion Support

shtukas:
This change adds bastion support to ssm. Essentially we introduce three new command line arguments `--bastion <aws instance id>`, `--bastion-port <number>` and `--bastion-user <username>`. The implementation assumes that the bastion doesn't prevent ssh agent forwarding.


src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -22,8 +22,16 @@ object Main {
...
               case _ => fail()
             }
-          case SsmSsh =>
-            setUpSSH(awsClients, executionTarget, user, sism, usePrivate, rawOutput)
+          case SsmSsh => {
+            val fProgramResult = bastionInstanceIdOpt match {
+              case None => setUpSSHStandardFutureProgramResult(awsClients, executionTarget, user, sism, usePrivate, rawOutput)
+              case Some(bastionInstanceId) => setUpSSHBastionFutureProgramResult(awsClients, executionTarget, user, sism, usePrivate, rawOutput, bastionInstanceId, bastionPortNumberOpt)
+            }
+            val programResult = Await.result(fProgramResult.asFuture, maximumWaitTime)
```
adamnfish: Why do we do this here now? In all other cases the actual execution (including the display) is done by the subprogram. It would make sense to do the same here? This change has just inlined the previous intermediate function (I forget the name, but maybe you know what I mean)? Better to abolish it?

I'd expect this to look like:

```scala
bastionInstanceIdOpt match {
  case None => standardSSH( ... )
  case Some(bastionInstanceId) => bastionSSH( ... )
}
```

and have those programs decide what their output should be.

adamnfish: Reference:
https://github.com/guardian/ssm-scala/pull/51/files#diff-fc8e9743688e677384a3956770151becR85

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,11 +53,26 @@ object Main {
...
-    programResult.fold(UI.outputFailure, UI.sshOutput(rawOutput))
-    System.exit(programResult.fold(_.exitCode, _ => 0))
+  private def setUpSSHBastionFutureProgramResult(awsClients: AWSClients, executionTarget: ExecutionTarget, user: String, sism: SingleInstanceSelectionMode, usePrivate: Boolean, rawOutput: Boolean, bastionInstanceId: String, bastionPortNumberOpt: Option[Int]) = {
+    for {
+      sshArtifacts <- Attempt.fromEither(SSH.createKey())
+      (privateKeyFile, publicKey) = sshArtifacts
+      bastionConfig <- IO.getSSMConfig(awsClients.ec2Client, awsClients.stsClient, ExecutionTarget(Some(List(InstanceId(bastionInstanceId)))))
+      bastionInstance <- Attempt.fromEither(Logic.getSSHInstance(bastionConfig.targets, SismUnspecified, false))
+      bastionAddAndRemovePublicKeyCommand = SSH.addTaintedCommand(bastionConfig.name) + SSH.addPublicKeyCommand(user, publicKey) + SSH.removePublicKeyCommand(user, publicKey)
+      bastionAddress <- Attempt.fromEither(Logic.getAddress(bastionInstance, false))
```
adamnfish: Why are these hard-coded to use a public address? It's plausible that the bastion could be on a peered network with a private IP address?

No worries for now if it addresses the MVP use case but we should keep in mind that this probably isn't right.

(It does make sense to hard code the target instance to use private IP addresses, so just here at the edge that the user should have the ability to switch between both)

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -73,4 +75,25 @@ object SSH {
     }
     (instance.id, cmd)
   }
+
+  def sshCmdBastion(rawOutput: Boolean)(privateKeyFile: File, bastionInstance: Instance, targetInstance: Instance, user: String, bastionIpAddress: String, targetIpAddress: String, bastionPortNumberOpt: Option[Int]): (InstanceId, String) = {
+    val portSpecifications = bastionPortNumberOpt match {
+      case Some(portNumber) => s"-p ${portNumber} " // trailing space is important
```
adamnfish: if it's important, maybe add a test for it?

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -45,11 +53,26 @@ object Main {
...
+      _ <- IO.tagAsTainted(bastionInstance.id, bastionConfig.name, awsClients.ec2Client)
```
adamnfish: I think we should do all the IO at once, after we've resolved everything that might fail. This means we install the keys and taint both the bastion and the target at the same time, once we know everything else has worked. Otherwise ot's likely we resolve the bastion, do the IO to taint it, and only then realise that the target can't be resolved. At that point we've done all that IO (and tainting) for no reason.

adamnfish: Also, I'm not 100% sure we need to taint a bastion, because that's literally its purpose. It means the motd would get pretty massive. We still get the audit trail from ec2 run command's logs and from cloudtrail.

This will be worth discussing.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -7,6 +7,8 @@ import java.util.Calendar
 import com.gu.ssm.utils.attempt._
 import com.gu.ssm.utils.{KeyMaker, FilePermissions}
 
+import scala.io.Source
```
adamnfish: I don't think this is used?

src/main/scala/com/gu/ssm/Main.scala
shtukas: You are perfectly right. I wasn't thinking straight and violated the principle that main just calls programs. It is now this:

```
case SsmSsh => bastionInstanceIdOpt match {
    case None => setUpStandardSSH(awsClients, executionTarget, user, sism, usePrivate, rawOutput)
    case Some(bastionInstanceId) => setUpBastionSSH(awsClients, executionTarget, user, sism, usePrivate, rawOutput, bastionInstanceId, bastionPortNumberOpt)
}
```

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Done!

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Left over from code that is now gone.

shtukas:
Until now the bastion user was assumed to be the same as the target instance user. I have just added support for customised bastion username: `--bastion-user <username>`.

shtukas:
> I like that it can re-use what exists, doesn't add much despite being a great new feature.

By the way @adamnfish, what you highlight is the result of one simple thing: the code was not only originally well designed, but remained, version after version, true to its original design principles [1].  

[1] That the people who came after you got a proper introduction to those ideas (even if I still sometimes accidentally violate them) did help :)

shtukas:
@jfsoul Going live :)

PR #58, opened 2018-04-06
More comprensive tests and fix a problem with the user command string in the bastion case

shtukas:


shtukas:
@adamnfish @aug24 FYI

shtukas:
+1

PR #61, opened 2018-04-19
Add -x or --execute option to immediately execute

aug24:
## What does this change?

Permits immediate execution of the ssh command so that the entire process is a single command.

## What is the value of this?

Simpler interface for users.

## Any additional notes?


shtukas:
Perfect, but I will add a commit to the branch before merging: README update

shtukas:
@aug24 Done. Merge at will.

PR #62, opened 2018-04-19
Bump version

aug24:
## What does this change?

Bumps version

## What is the value of this?

Bumps version

## Any additional notes?

No


shtukas:
@aug24 This is one of those case where a review before merging was required ;)

PR #64, opened 2018-04-19
Unbump version pending further work to support Discussion team

aug24:
## What does this change?

Rebump version to 0.9.1 prior to release for homebrew.

## What is the value of this?

Release including --execute without marking as v1

## Any additional notes?


shtukas:
@aug24 

+1, if comes after https://github.com/guardian/ssm-scala/pull/63 , and commit message is updated with "bump", instead of "un-bump" and if it doesn't mention Discussion :)

PR #66, opened 2018-04-23
Introduces the -a, --agent flags

shtukas:
There's been occurences of bastions connections strings of the form

```
ssh -A -i /path/to/temp/private/key -t -t ubuntu@bastion-hostname \
    -t -t ssh -t -t ubuntu@target-ip-address;
```

not working, because the private file was not found for the second ssh connection, leading to a "Permission denied (publickey)" error message.

When this happens the user can use the `-a`, `--agent` flag that performs a registration of the private key at the local ssh agent. With this flag, ssm command

```
ssm ssh --profile <account-name> --bastion <instanceId1> \
    -i <instanceId2> --agent
```

returns

```
ssh-add /path/to/temp/private/key && \
    ssh -A ubuntu@bastion-hostname \
    -t -t ssh ubuntu@target-ip-address;
```

### DIFF
--- a/src/main/scala/com/gu/ssm/SSH.scala
+++ b/src/main/scala/com/gu/ssm/SSH.scala
@@ -90,8 +90,8 @@ object SSH {
     } else {
       s"ssh -A${bastionPortSpecifications}${stringFragmentMinusIOption}${stringFragmentTTOptions} $bastionUser@$bastionIpAddress"
     }
-    val stringFragmentTargetConnection = s"ssh${targetPortSpecifications}${stringFragmentTTOptions} $targetInstanceUser@$targetIpAddress"
-    val connectionString = s"${stringFragmentSshAdd}${stringFragmentBastionConnection} -t -t ${stringFragmentTargetConnection}"
+    val stringFragmentTargetConnection = s"-t -t ssh${targetPortSpecifications}${stringFragmentTTOptions} $targetInstanceUser@$targetIpAddress"
+    val connectionString = s"${stringFragmentSshAdd}${stringFragmentBastionConnection} ${stringFragmentTargetConnection}"
     val cmd = if(rawOutput) {
       s"$connectionString"
     }else{
--- a/src/main/scala/com/gu/ssm/SSH.scala
+++ b/src/main/scala/com/gu/ssm/SSH.scala
@@ -79,19 +79,19 @@ object SSH {
     (instance.id, cmd)
   }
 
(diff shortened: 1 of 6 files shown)


QuarpT:
I tested this locally and it works.

There's no clean up for temporary identities added to the authentication agent although this isn't a big issue.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -79,19 +79,19 @@ object SSH {
...
-    val targetPortSpecifications = targetInstancePortNumberOpt match {
-      case Some(portNumber) => s" -p ${portNumber}" // trailing space is important
-      case _ => ""
-    }
-    val theTTOptions = if(rawOutput) { " -t -t" }else{ "" }
-    val connectionString1 = s"ssh -A${bastionPortSpecifications} -i ${privateKeyFile.getCanonicalFile.toString}${theTTOptions} $bastionUser@$bastionIpAddress"
-    val connectionString2 = s"ssh${targetPortSpecifications}${theTTOptions} $targetInstanceUser@$targetIpAddress"
-    val connectionString = s"${connectionString1} -t -t ${connectionString2}"
+    val stringFragmentTargetConnection = s"ssh${targetPortSpecifications}${stringFragmentTTOptions} $targetInstanceUser@$targetIpAddress"
+    val connectionString = s"${stringFragmentSshAdd}${stringFragmentBastionConnection} -t -t ${stringFragmentTargetConnection}"
```
aug24: Do we want -t-t here, or should this be the variable?

shtukas:
@QuarpT 

> no clean up for temporary identities added to the authentication agent

Yep, I will address that in a separate PR :)

src/main/scala/com/gu/ssm/SSH.scala
shtukas: I make a difference between the `-t -t` carried by `stringFragmentTTOptions` which only comes for raw outputs, and the one in the final connection string which is always there.

shtukas: But since my understanding of what it does is limited, feel free to correct me and I will update the code accordingly :)

aug24: If it's always there, then *I* would put it in the stringFragmentTargetConnection string.  Just so that it's clear where/why.  So long as it is intended though, I'm happy.  I'll approve in a sec.

PR #70, opened 2018-05-01
Ensure correct ownership of the authorized_keys file on target

shtukas:
On some instances authorized_keys comes owned by root. This change ensures that the file is owned by ubuntu or the default user provided by the user.


tomrf1:
👍 thanks @shtukas

PR #80, opened 2018-05-24
Introduce new sub command 'scp' for secure copy

shtukas:
This PR adds the **scp** sub command to **ssm**, for performing secure copy.

## Secure Copy

**ssm** support the **scp** sub command for the secure transfer of files and directories.

### Introduction

An example of usage is 

```
./ssm scp -p account -t app,stack,stage /path/to/file1 :/path/to/file1
``` 

Which outputs

```
# simplified version
scp -i /path/to/identity/file.tmp /path/to/file1 ubuntu@34.242.32.40:/path/to/file2;
```

Otherwise 

```
./ssm scp -p account -t app,stack,stage :/path/to/file1 /path/to/file2
```

outputs 

```
# simplified version
scp -i /path/to/identity/file.tmp ubuntu@34.242.32.40:/path/to/file1 /path/to/file2 ;
```

For the moment we only support moving files. Support for directories will come in the next PR.

### DIFF
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -154,6 +154,57 @@ object ArgumentParser {
           else success )
       )
 
+    cmd("scp")
+      .action((_, c) => c.copy(mode = Some(SsmScp)))
+      .text("Secure Copy")
+      .children(
+        opt[String]('u', "user").optional()
+          .action((user, args) => args.copy(targetInstanceUser = Some(user)))
+          .text(s"Connect to remote host as this user (default: $targetInstanceDefaultUser)"),
+        opt[Int]("port").optional()
+          .action((port, args) => args.copy(targetInstancePortNumber = Some(port)))
+          .text(s"Connect to remote host on this port"),
+        opt[Unit]("newest").optional()
+          .action((_, args) => {
+            args.copy(
+              singleInstanceSelectionMode = SismNewest,
+              isSelectionModeNewest = true)
+          })
+          .text("Selects the newest instance if more than one instance was specified"),
+        opt[Unit]("oldest").optional()
+          .action((_, args) => {
+            args.copy(
+              singleInstanceSelectionMode = SismOldest,
+              isSelectionModeOldest = true)
+          })
+          .text("Selects the oldest instance if more than one instance was specified"),
+        opt[Unit]("private").optional()
+          .action((_, args) => {
+            args.copy(
+              usePrivateIpAddress = true)
+          })
+          .text("Use private IP address (must be routable via VPN Gateway)"),
+        opt[Unit]("raw").optional()
+          .action((_, args) => {
+            args.copy(
+              rawOutput = true)
+          })
+          .text("Unix pipe-able scp connection string"),
+        opt[Unit]('x', "execute").optional()
+          .action((_, args) => {
+            args.copy(
+              rawOutput = true)
+          })
+          .text("Makes ssm behave like a single command (eg: `--raw` with automatic piping to the shell)"),
+        arg[String]("<sourceFile>...").required()
+          .action( (sourceFile, args) => args.copy(sourceFile = Some(sourceFile)) ),
+        arg[String]("<targetFile>...").required()
+          .action( (targetFile, args) => args.copy(targetFile = Some(targetFile)) ),
+        checkConfig( c =>
+          if (c.isSelectionModeOldest && c.isSelectionModeNewest) failure("You cannot both specify --newest and --oldest")
+          else success )
+      )
+
     checkConfig { args =>
       if (args.mode.isEmpty) Left("You must select a mode to use: cmd, repl or ssh")
       else if (args.toExecute.isEmpty && args.mode.contains(SsmCmd)) Left("You must provide commands to execute (src-file or cmd)")
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -197,9 +197,11 @@ object ArgumentParser {
           })
           .text("Makes ssm behave like a single command (eg: `--raw` with automatic piping to the shell)"),
         arg[String]("<sourceFile>...").required()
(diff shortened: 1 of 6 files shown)


src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -151,16 +159,28 @@ object SSH {
       case None => ""
       case Some(decision) => if(decision) " -A" else " -a"
     }
-    val connectionString = s"""scp -o "IdentitiesOnly yes"$useAgentFragment$hostsFileString${targetPortSpecifications} -i ${privateKeyFile.getCanonicalFile.toString}${theTTOptions} ${sourceFile} $user@$ipAddress:${targetFile}"""
-    val cmd = if(rawOutput) {
-      s"$connectionString"
+    // We are using colon to designate the remote file.
+    // There should be only one.
+    if (startingColonToInt(sourceFile)+startingColonToInt(targetFile) == 1) {
```
sihil: Shame there is not a logical XOR operator in Scala. This isn't super clear to me. Perhaps the function should be called `isRemote`?

Is something like this any better? : `List(sourceFile, targetFile).map(isRemote).count(_ == true)`

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -140,6 +140,14 @@ object SSH {
   // The first file goes to the second file
   // The remote file is indicated by a colon
 
+  def startingColonToInt(string: String): Integer = {
+    if(string.startsWith(":")) 1 else  0
+  }
+
+  def removeFirstLetter(string: String): String = {
+    string.substring(1)
```
sihil: Maybe use `string.stripPrefix(':')` and then you can pass both through it without caring?

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -137,4 +137,50 @@ object SSH {
...
+    val useAgentFragment = useAgent match {
+      case None => ""
+      case Some(decision) => if(decision) " -A" else " -a"
+    }
+    // We are using colon to designate the remote file.
+    // There should be only one.
+    if (startingColonToInt(sourceFile)+startingColonToInt(targetFile) == 1) {
+      val connectionString =
+        if (startingColonToInt(sourceFile)==1) {
+          s"""scp -o "IdentitiesOnly yes"$useAgentFragment$hostsFileString${targetPortSpecifications} -i ${privateKeyFile.getCanonicalFile.toString}${theTTOptions} $user@$ipAddress:${removeFirstLetter(sourceFile)} ${targetFile}"""
```
sihil: We should extract all these options out somewhere as they are common to both `ssh` and `scp`. Possibly something to do when we migrate to a config file.

src/main/scala/com/gu/ssm/ArgumentParser.scala
```diff
@@ -154,6 +154,59 @@ object ArgumentParser {
...
+              rawOutput = true)
+          })
+          .text("Unix pipe-able scp connection string"),
+        opt[Unit]('x', "execute").optional()
+          .action((_, args) => {
+            args.copy(
+              rawOutput = true)
+          })
+          .text("Makes ssm behave like a single command (eg: `--raw` with automatic piping to the shell)"),
+        arg[String]("<sourceFile>...").required()
```
sihil: With scp you can specify multiple source files. Would be nice to support this.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -137,4 +137,50 @@ object SSH {
     (targetInstance.id, cmd)
   }
 
+  // The first file goes to the second file
+  // The remote file is indicated by a colon
+
+  def startingColonToInt(string: String): Integer = {
```
sihil: isRemote and return boolean?

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -137,4 +137,50 @@ object SSH {
...
+    }
+    val hostsFileString = hostsFile.map(file => s""" -o "UserKnownHostsFile $file" -o "StrictHostKeyChecking yes"""").getOrElse("")
+    val theTTOptions = if(rawOutput) { " -t -t" }else{ "" }
+    val useAgentFragment = useAgent match {
+      case None => ""
+      case Some(decision) => if(decision) " -A" else " -a"
+    }
+    // We are using colon to designate the remote file.
+    // There should be only one.
+    if (startingColonToInt(sourceFile)+startingColonToInt(targetFile) == 1) {
```
sihil: You could just let scp itself deal with incorrect file lists rather than reimplementing the logic here.

For example... if you have a list of files you could choose to do something like: `files.map(expandFileParam(s"$user@$ipAddress", _))` 

with an `expandFileParam` that does something like:

```
def expandFileParam(host: String, file: String): String = {
  if (isRemote(file)) s"$host$file" else file
}
```

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Done. (Refactored to make it much more clear.)

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Done. I didn't know `.stripPrefix(":")`.

src/main/scala/com/gu/ssm/SSH.scala
shtukas: https://github.com/guardian/ssm-scala/issues/99

src/main/scala/com/gu/ssm/ArgumentParser.scala
shtukas: Will do. Together when I add support for `-r`

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Fixed in refactoring

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Solved in refactoring, but recorded idea for future reference :)

PR #81, opened 2018-06-06
Add bastion-tags option

aug24:
## What does this change?

Allows the user to specify the bastion by its tags.

## What is the value of this?

It removes the need to embed marauder sub-shell commands in the ssm/ssh command.


shtukas:
Let me know if you want me to release a new version after you merge.

nicl:
👍 for the functionality as this would be really useful for Discussion too :)

PR #84, opened 2018-06-08
Improve management of user and host keys

### LINKED ISSUE #73: When SSHing get host key from server and enforce use in SSH command
At the moment the user must accept the remote host's SSH key when they connect. This should be transparently verified without a user having to deal with it.

The SSM invocation could and should fetch the remote host key (probably `/etc/ssh/ssh_host_ecdsa_key.pub`) and write it to a temporary local host keys file in the correct format (`hostname key_format key`).

The resulting ssh command should then use `UserKnownHostsFile` (to specify the file) and `StrictHostKeyChecking=yes` (`ask` is default, `yes` refuses to connect if the key isn't known) to point at the temporary file and refuse to connect if the host file is not known. e.g.:
```
ssh -o UserKnownHostsFile=/my/tmp/file -o StrictHostKeyChecking=yes <host> etc etc ...
```

This will provide stronger security guarantees as the user will definitely be connecting to the server that SSM has talked to the agent on rather than any MitM.

/cc @adamnfish @shtukas @aug24

### LINKED ISSUE #79: "Too many authentication failures"
# ISSUE
If you have a lot of keys added to your agent ssm can fail with "Too many authentication failures".

By adding `-o 'IdentitiesOnly yes'` to the ssh command line this issue should go away as it will not use your agent.

sihil:
## What does this change?

This PR fixes #73 by changing the async command made via SSM into a synchronous command that adds the public key and retrieves the host keys on the remote machine. These host keys are examined locally to find the preferred key, which is then saved to a file and passed to the SSH command in order to verify the remote host. In the case of the bastion hosts, both the host key from the bastion and the host key from the target host are added to the local host keys file.

To simplify the key management, the bastion host work has been changed to use `ProxyCommand`. This bypasses the need to use ssh-add to temporarily add keys to the agent as the identity can be provided directly to both the bastion and target hosts.

In addition this PR adds `-o 'IdentitiesOnly yes'` to all SSH invocations which fixes #79. 

## What is the value of this?

The main advantage of this is users will no longer have to type `yes` to accept a host key for a server they have not accessed before. In addition the `WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!` error will no longer occur as we stop relying on the user's default host key file.

## Any additional notes?

This improves security by always validating the host key retrieved via SSM, which eliminates the risk of MitM attacks. To prevent users bypassing this we also add `StrictHostKeyChecking yes` which prevents users being prompted to add unknown host keys.

A (positive, I think) side effect of this change is that the SSH process is less racy as it now waits for the public key to be put on the server.


shtukas:
Nice one!

shtukas:
@sihil 
I will look at it carefully, and do manual testings beyond the automated testing, and will come back to you :)

src/test/scala/com/gu/ssm/utils/attempt/AttemptTest.scala
```diff
@@ -44,7 +44,7 @@ class AttemptTest extends FreeSpec with Matchers with EitherValues with AttemptV
   }
 
   "delay" - {
-    "will cause timeout in shorter time" in {
+    "will cause timeout in shorter time" ignore {
```
shtukas: Thank you, this was driving me crazy at times.

shtukas:
@sihil 
Merge at will, I have checked the simple as well as bastion connections against live instances. All good as far as I can see.

Issue #74, opened 2018-05-16
--private should be automatic where possible


sihil:
# ISSUE
When sshing to a machine over direct connect `--private` should be applied automatically (especially when the instance has no public IP!)

## Steps to Reproduce
1. ssh to a machine with no public IP

## Actual Results (include screenshots)
If you ssh to a machine without a public IP the command refuses, saying `Found i-07af7c2fc6fd32fc3 but no instances have public IPs (use '--private' if internal address is routable)`.

## Expected Results (include screenshots)
ssm should try to connect to the private IP directly, and _assume_ that it is routable (a socket connect on 22 would be pessimistic but potentially avoid unhelpful timeouts).

shtukas:
@sihil Glancing over the code, that's a easy enough change to do. I actually propose to just drop the `--private ` flag and automatically use the private IP when there isn't a public one. Thoughts ?

shtukas:
Talked with @sihil and just to clarify: "a socket connect on 22 would be pessimistic but potentially avoid unhelpful timeouts" means from the scala code with a short timeout and in the failing case a message to the user saying that the instance has no public address, has one private address, but that private address is not routable. 

In which case, dropping  `--private` is, indeed, the right decision.

shtukas:
Picking this up now. I am not going to drop `--private` (no need to break existing functionalities), but will simply imply it when relevant. The idea around trying to reach port 22 before reporting no routability will come in a different PR.

shtukas:
Addressed in: https://github.com/guardian/ssm-scala/pull/86

PR #86, opened 2018-06-12
Unless --private, use public IP when available and otherwise private IP

shtukas:
## What does this change?

When the `--private` flag is used, we use instances private IPs. If not, we default to the public IP when available and otherwise use the private IP.

Little detail: moving from

```
def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode, usePrivate: Boolean): Either[FailedAttempt, Instance]
```

to 

```
def getSSHInstance(instances: List[Instance], sism: SingleInstanceSelectionMode): Either[FailedAttempt, Instance]
```

simplified the tests.

## What is the value of this?

Simpler user interface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,7 @@ An example of using `repl` is:
 ./ssm repl --profile <aws-profile> -t security-hq,security,PROD
 ```
 
-The REPL mode causes `ssm` to generate a list of
-instances and then wait for commands to be specified.  Each command will be executed on all instances and the user can select the instance to display.
+The REPL mode causes `ssm` to generate a list of instances and then wait for commands to be specified. Each command will be executed on all instances and the user can select the instance to display.
 
 An example of using `ssh` command is:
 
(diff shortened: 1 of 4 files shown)


shtukas:
Address: https://github.com/guardian/ssm-scala/issues/74

shtukas:
@sihil @aug24 Worth paying attention to the tests. They got simplified.

JonNorman:
I've found that when SSH'ing onto EMR clusters, despite them having a public and private address, only the latter works. I think the DNS lookup fails for the public one though I haven't sought to understand why.

Would it be possible to retain this flag to force it to use the private address if both are available?

shtukas:
@JonNorman 
This is **exactly** the feedback I wanted. I will update then.

shtukas:
@JonNorman Done.

shtukas:
In any case, I am happy with the flag staying in :)

JonNorman:
@sihil yeah I'd prefer to do that but having had a bit of a route around it's not obvious where to specify this... will look again when I've a bit more time.

aug24:
Perfectly possible to have public and private IPs and only one of them accepts requests for a specific port.  

I think we should default to: public, or private if public doesn't exist.

But we must have --private to force the issue if both exist and public ip doesn't accept ssh.

shtukas:
@aug24 

Simon had this idea (which is mentioned in the related issue https://github.com/guardian/ssm-scala/issues/74 ), that we could actually synchronously try connecting to the target ssh port before giving a connection string back to the user. 

This solution gracefully handles edge cases as the one you mention, but there was no way I would have overloaded one PR with that :)

PR #89, opened 2018-06-19
Introduce flag --identity-file

shtukas:
## What does this change?
Introduce flag `--identity-file` causing only the identity file path to be displayed (for use in scripts)

## What is the value of this?
Allow custom shell scripts built around ssm to be much more robust.

### DIFF
--- a/src/main/scala/com/gu/ssm/SSH.scala
+++ b/src/main/scala/com/gu/ssm/SSH.scala
@@ -90,7 +90,7 @@ object SSH {
        | for hostkey in $$(grep ^HostKey $sshd_config_path| cut -d' ' -f 2); do cat $${hostkey}.pub; done
      """.stripMargin
 
-  def sshCmdStandard(rawOutput: Boolean, shouldDisplayIdentityFileOnly: Boolean)(privateKeyFile: File, instance: Instance, user: String, ipAddress: String, targetInstancePortNumberOpt: Option[Int], hostsFile: Option[File], useAgent: Option[Boolean]): (InstanceId, String) = {
+  def sshCmdStandard(rawOutput: Boolean, shouldDisplayIdentityFileOnly: Boolean, privateKeyFile: File, instance: Instance, user: String, ipAddress: String, targetInstancePortNumberOpt: Option[Int], hostsFile: Option[File], useAgent: Option[Boolean]): (InstanceId, String) = {
     val targetPortSpecifications = targetInstancePortNumberOpt match {
       case Some(portNumber) => s" -p $portNumber" // trailing space is important
       case _ => ""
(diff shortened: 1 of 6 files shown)


shtukas:
Originated by the fact that this PR (caused by a recent update in ssm) https://github.com/guardian/discussion-platform/pull/333 , should not have occurred. 

Note that once this updated ssm goes live, the discussion scripts will be updated.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -102,8 +102,12 @@ object SSH {
     }
     val hostsFileString = hostsFile.map(file => s""" -o "UserKnownHostsFile $file" -o "StrictHostKeyChecking yes"""").getOrElse("")
     val connectionString = s"""ssh -o "IdentitiesOnly yes"$useAgentFragment$hostsFileString$targetPortSpecifications -i ${privateKeyFile.getCanonicalFile.toString}${theTTOptions} $user@$ipAddress"""
-    val cmd = if(rawOutput) {
-      s"$connectionString"
+    val cmd = if(rawOutput || shouldDisplayIdentityFileOnly) {
```
nicl: Might be simpler as:

    if (shouldDisplayIdentityFileOnly) {
      ..
    } else if (rawOutput) {
      ..
    } else ...

?

shtukas: Updated, pushed, thanks :)

shtukas:
@nicl Thanks. I won't merge before @sihil approve :)

sihil:
This side steps all of the helpful HostKey validation that SSM does. This is highlighted by the fact that the script in discussion has to do `-o StrictHostKeyChecking=no` - as a result there is no guarantee that you have connected to the server that ssm told you about.

I wonder if a better solution would be to have an option that outputs all of the helpful ssh parameters for you - in either a format for ssh or scp as appropriate. This should include `-i`, `IdentitiesOnly`, `UserKnownHostsFile` and `StrictHostKeyChecking` (set to yes). 

On a side note, now that SSM has been changed to use `ProxyCommand` rather than nested SSH sessions when dealing with bastions, setting up an SSH tunnel should be far easier than your script is doing and I'm not sure you need to chop up the ssm command at all - you should just be able to do `$(ssm --raw) -L ...`.

sihil:
I've just spoken with @nicl. We proved that we can in theory massively simplify the script in discussion and that this `--identity-file` is not needed. Unfortunately we also discovered that `--raw` is broken (see #90).

I'd advocate for this PR to be closed as it provides a path that weakens security when other compelling routes are available.

shtukas:
> We proved that we can in theory massively simplify the script in discussion 

Perfect. I am closing this then :)

PR #91, opened 2018-06-20
Improve the documentation for the --raw option

sihil:
## What does this change?
My recent changes to SSM in #84 made liberal use of quotes when specifying options (particularly around `ProxyCommand` for bastion access). This breaks `--raw` if you execute it directly as bash will interpret `ssh -o "IdentitiesOnly yes"` as four separate terms: `ssh`, `-o`, `"IdentitiesOnly` and `yes"`. By using `eval` it ensures that it is evaluated as if it had been typed into a terminal directly.

This is therefore a documentation change that advises people to use eval in combination with `--raw`

## What is the value of this?
People were getting an error when using `--raw`. This documents how it can be used.

## Any additional notes?
No

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Create and upload a temporary ssh key
   --newest                 Selects the newest instance if more than one instance was specified
   --oldest                 Selects the oldest instance if more than one instance was specified
   --private                Use private IP address (must be routable via VPN Gateway)
-  --raw                    Unix pipe-able ssh connection string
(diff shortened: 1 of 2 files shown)


shtukas:
Thanks @sihil

sihil:
No worries. This isn't really changing much so happy for it to go out whenever there is something useful.

shtukas:
I will merge so that the online readme is updated, but I would not do anything else :)

Issue #92, opened 2018-06-20
ssm should not write to /tmp/ssm.log


shtukas:
@rtyley reported that on a machine with more than one user account (for instance the machine he shares with Amy), the fact that the log file is shared ( see: https://github.com/guardian/ssm-scala/blob/master/src/main/resources/logback.xml ) causes problems (essentially ssm breaks for the _second_ user). Should be an easy fix. This log file should be written in the user's home directory, for instance `~/.ssm/ssm.log`.

shtukas:
Solved by: https://github.com/guardian/ssm-scala/pull/93

Issue #1, opened 2018-01-12
Logging


adamnfish:
The logger should be off by default, and on (to stdout) when `--debug` is provided. Sadly `slf4j` (and thus `scala-logging`) does not support setting the log level at runtime.

TODO: Work out the incantations required to obtain the underlying logback-classic logger and set its log level based on the presence (/absence) of a `--debug` flag.

Related: Scatter logging all through the application.
Related: The --debug flag could optionally take a level if we want to take full advantage of the logger's power.

PR #93, opened 2018-06-20
Move the log file to the user's home directory

shtukas:
Solves: https://github.com/guardian/ssm-scala/issues/92

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -1,6 +1,6 @@
 <configuration>
     <appender name="FILE" class="ch.qos.logback.core.FileAppender">
-        <file>/tmp/ssm.log</file>
+        <file>${user.home}/.ssm/ssm.log</file>
         <append>false</append>
         <encoder>
             <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>


src/main/resources/logback.xml
```diff
@@ -1,6 +1,6 @@
 <configuration>
     <appender name="FILE" class="ch.qos.logback.core.FileAppender">
-        <file>/tmp/ssm.log</file>
+        <file>${user.home}/.ssm/ssm.log</file>
```
rtyley: Sounds legit- does it create the `~/.ssm/` folder if it doesn't exist?

sihil:
Out of interest, do we really need a log file?

shtukas:
Um... I would think it's a relic of early stages of ssm.  @aug24 @adamnfish Thoughts ?

src/main/resources/logback.xml
shtukas: Yes, it does. I tested it :)

adamnfish:
I don't think there's actually much logging happening at the moment. Logging is nice though, it was very helpful during the early development phase.

This was the plan
https://github.com/guardian/ssm-scala/issues/1

Obviously, I don't know how to do that...

shtukas:
Thanks Adam. I am merging this one for the moment.

sihil:
`-v` when you need it, gets output to stderr?

Issue #94, opened 2018-06-21
Temporary key files are never cleaned up


rahilb:
# ISSUE

I have lots of keys in my temp directory after a few days of using ssm

## Steps to Reproduce

Use ssm for a bit, list files in your temp dir

## Actual Results (include screenshots)

```
 rahil_bohra@32597  /var/folders/_r/0119p6z94p1d7mktzp05g0n1b00456/T> ll | grep ssm
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   398B 18 Jun 11:33 security_ssm-scala_temporary-host-key125807685762587641.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   398B 18 Jun 15:28 security_ssm-scala_temporary-host-key1410296220664484847.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   398B 18 Jun 10:45 security_ssm-scala_temporary-host-key1792334240323237599.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 20 Jun 12:12 security_ssm-scala_temporary-host-key1825282419528675198.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 20 Jun 12:06 security_ssm-scala_temporary-host-key2243716221987320095.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 20 Jun 12:01 security_ssm-scala_temporary-host-key2699408009068729302.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   396B 18 Jun 15:13 security_ssm-scala_temporary-host-key3056040440274860416.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 19 Jun 13:39 security_ssm-scala_temporary-host-key3206763035813787855.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 20 Jun 11:24 security_ssm-scala_temporary-host-key3971101146287793205.tmp
-rw-r--r--   1 rahil_bohra  GNM\WG GNM Unmanaged   393B 20 Jun 12:01 security_ssm-scala_temporary-host-key4030082673690199416.tmp
[...]

PR #95, opened 2018-06-21
Moving user's temporary files to OS's temp folder

shtukas:
The temporary files generated by ssm are not automatically cleaned up, so we are moving them to the OS's temp folder.

### DIFF
--- a/src/main/scala/com/gu/ssm/SSH.scala
+++ b/src/main/scala/com/gu/ssm/SSH.scala
@@ -20,7 +20,7 @@ object SSH {
     val keyProvider = "BC"
 
     try {
-      val privateKeyFile = File.createTempFile(prefix, suffix)
+      val privateKeyFile = File.createTempFile(prefix, suffix, new File(System.getProperty("java.io.tmpdir")))
       FilePermissions(privateKeyFile, "0600")
       val publicKey = KeyMaker.makeKey(privateKeyFile, keyAlgorithm, keyProvider)
       Right((privateKeyFile, publicKey))


src/main/scala/com/gu/ssm/UserFiles.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.ssm
+
+import java.io.File
+
+object UserFiles {
+  val userHomeDirectory = System.getProperty("user.home")
+  val userSMMFolder = s"${userHomeDirectory}/.ssm"
```
rahilb: `File.separator` should be used, in case we'd ever like to support windows machines.

rahilb: better still, `Paths.get(userHomeDirectory, ".ssm")`

shtukas:
@sihil I based the decision on Adam's input 
https://github.com/guardian/ssm-scala/issues/94#issuecomment-399124475

src/main/scala/com/gu/ssm/UserFiles.scala
shtukas: @sihil 
Back in the days, we decided we would be portable for window users. So @rahilb is sort of relevant. Update coming.

sihil:
Sure, but @adamnfish's comment was based on not being able to get a common tmp dir.

shtukas:
Oh!
Um..... yeah. So I guess that you're saying that using the existing `/tmp` and removing files from there is _better_ than using our own `~/.ssm/tmp` (and removing files from there). It's all the same to me. So coming update then :)

shtukas:
@sihil @rahilb Ok, ready. Note that technically I didn't need to create `UserFiles.scala`, but this is where i will put the cleaning function.

shtukas:
@sihil @rahilb Done

adamnfish:
In the OS' temporary directory the files will be automatically removed for us, is why this is better. There's now no need to add a feature to `ssm` that will sweep that dir.

I've checked on Linux and this correctly creates a temporary file under `/tmp` there, assume this is the same on OSX as well, nice one!

PR #102, opened 2018-08-08
Move IO right to the edge, allow flexible output and add --verbose flag

sihil:
This replaces the string output type with various output types that can be processed by the UI classes. By allowing this to be returned by the actual methods we grant much greater flexibility in how output is presented (and can be optionally presented using the new `--verbose` flag).

### DIFF
--- a/src/main/scala/com/gu/ssm/SSH.scala
+++ b/src/main/scala/com/gu/ssm/SSH.scala
@@ -26,13 +26,13 @@ object SSH {
       Right((privateKeyFile, publicKey))
     } catch {
       case e:IOException => Left(FailedAttempt(
-        Failure(s"Unable to create private key file", "Error creating key on disk", UnhandledError, None, Some(e))
+        Failure(s"Unable to create private key file", "Error creating key on disk", UnhandledError, e)
       ))
       case e:NoSuchAlgorithmException => Left(FailedAttempt(
-        Failure(s"Unable to create key pair with algorithm $keyAlgorithm", s"Error creating key with algorithm $keyAlgorithm", UnhandledError, None, Some(e))
+        Failure(s"Unable to create key pair with algorithm $keyAlgorithm", s"Error creating key with algorithm $keyAlgorithm", UnhandledError, e)
       ))
       case e:NoSuchProviderException => Left(FailedAttempt(
-        Failure(s"Unable to create key pair with provider $keyProvider", s"Error creating key with provider $keyProvider", UnhandledError, None, Some(e))
+        Failure(s"Unable to create key pair with provider $keyProvider", s"Error creating key with provider $keyProvider", UnhandledError, e)
       ))
     }
   }
@@ -55,7 +55,7 @@ object SSH {
       Attempt.Right(hostKeyFile)
     } catch {
       case e:IOException => Attempt.Left(
-        Failure(s"Unable to create host key file", "Error creating host key on disk", UnhandledError, None, Some(e))
+        Failure(s"Unable to create host key file", "Error creating host key on disk", UnhandledError, e)
       )
     }
   }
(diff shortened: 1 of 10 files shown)


src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -102,18 +102,18 @@ object SSH {
...
     val hostsFileString = hostsFile.map(file => s""" -o "UserKnownHostsFile $file" -o "StrictHostKeyChecking yes"""").getOrElse("")
     val connectionString = s"""ssh -o "IdentitiesOnly yes"$useAgentFragment$hostsFileString$targetPortSpecifications -i ${privateKeyFile.getCanonicalFile.toString}${theTTOptions} $user@$ipAddress"""
-    val cmd = if(rawOutput) {
-      s"$connectionString"
-    }else{
-      s"""
-         | # Execute the following command within the next $sshCredentialsLifetimeSeconds seconds:
-         | ${connectionString};
-         |""".stripMargin
+    val cmd = if (rawOutput) {
```
LATaylor-guardian: This pattern is used in three places within this file - maybe you could pull out into a small function?

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -102,18 +102,18 @@ object SSH {
...
+      Seq(
+        Metadata(s"# Execute the following command within the next $sshCredentialsLifetimeSeconds seconds:"),
+        Out(s"$connectionString;")
+      )
     }
     (instance.id, cmd)
   }
 
-  def sshCmdBastion(rawOutput: Boolean)(privateKeyFile: File, bastionInstance: Instance, targetInstance: Instance, targetInstanceUser: String, bastionIpAddress: String, targetIpAddress: String, bastionPortNumberOpt: Option[Int], bastionUser: String, targetInstancePortNumberOpt: Option[Int], useAgent: Option[Boolean], hostsFile: Option[File]): (InstanceId, String) = {
+  def sshCmdBastion(rawOutput: Boolean)(privateKeyFile: File, bastionInstance: Instance, targetInstance: Instance, targetInstanceUser: String, bastionIpAddress: String, targetIpAddress: String, bastionPortNumberOpt: Option[Int], bastionUser: String, targetInstancePortNumberOpt: Option[Int], useAgent: Option[Boolean], hostsFile: Option[File]): (InstanceId, Seq[Output]) = {
```
LATaylor-guardian: Would be great to have some type safety around the String arguments to this function. Outside of this PR though.

src/main/scala/com/gu/ssm/SSH.scala
```diff
@@ -127,20 +127,20 @@ object SSH {
...
+      )
     }
     (targetInstance.id, cmd)
   }
 
   // The first file goes to the second file
   // The remote file is indicated by a colon
 
-  def scpCmdStandard(rawOutput: Boolean)(privateKeyFile: File, instance: Instance, user: String, ipAddress: String, targetInstancePortNumberOpt: Option[Int], useAgent: Option[Boolean], hostsFile: Option[File], sourceFile: String, targetFile: String): (InstanceId, String) = {
+  def scpCmdStandard(rawOutput: Boolean)(privateKeyFile: File, instance: Instance, user: String, ipAddress: String, targetInstancePortNumberOpt: Option[Int], useAgent: Option[Boolean], hostsFile: Option[File], sourceFile: String, targetFile: String): (InstanceId, Seq[Output]) = {
```
LATaylor-guardian: As above.

src/main/scala/com/gu/ssm/SSH.scala
shtukas: Good point

LATaylor-guardian:
👍

PR #108, opened 2019-01-25
Corrects raw mode equivalent examples

tonytw1:
Corrects raw mode equivalent examples to include ssm prefix which seems to be required.


shtukas:
Thanks :)

PR #109, opened 2019-06-06
Add support for -u option in `ssm cmd`

regiskuckaertz:
## What does this change?

It is sometimes necessary to execute remote bash commands as another user, e.g. when the bash script is in that user's home folder, but also to make sure proper permissions are set on generated files etc.

## What is the value of this?

The `-u` option is already available for `ssm ssh`, applying it to others seems to follow naturally.


shtukas:
May I suggest that you update the README as well. People look it up quite often :)

regiskuckaertz:
Ah yeah, good shout

PR #111, opened 2019-07-19
Optionally tunnel SSH through Systems Manager

mbarton:
## What does this change?

Adds `--ssm-tunnel` flag to [tunnel connections using AWS Systems Manager](https://aws.amazon.com/about-aws/whats-new/2019/07/session-manager-launches-tunneling-support-for-ssh-and-scp/). In practice this means you can access machines without the use of a bastion host and can close port 22 for access entirely, even within the VPC.

This requires the Systems Manager agent to be [at least 2.3.672.0](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html). ~Unfortunately this image is not yet in the Ubuntu base AMI images we use (not even the nightlies as of 19/07/19)~ The image is in the base images we use for AMIgo.

It also requires the instance policies to have additional `ssmmessages` [policy statements](https://docs.aws.amazon.com/systems-manager/latest/userguide/getting-started-create-iam-instance-profile.html):

```yaml
- "ssm:UpdateInstanceInformation"
- "ssmmessages:CreateControlChannel"
- "ssmmessages:CreateDataChannel"
- "ssmmessages:OpenControlChannel"
- "ssmmessages:OpenDataChannel"
```

Because of these requirements I have added it behind a flag. It might be possible to remove the flag by performing a look-up to see if the instance is running the newer agent and tunnelling, falling back to the traditional SSH directly. Naturally this would negate the security benefit from closing port 22.

I think it would also be possible to make a version of `ssm-scala` that is designed to be run via `ProxyCommand` in ssh config, following the trick [suggested in the AWS docs](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html). It would initially provision the temporary SSH key on the machine and then open a proxy to the `aws-cli` that then in turn proxies to the machine. I will investigate further as this would provide us with a cleaner migration path being a totally separate command, but I think it's important to have the capability up front, hence the PR here.

## What is the value of this?

Increased security posture by supporting auditable SSH connections to instances without exposing port 22 outside of each instance.

## Any additional notes?

### DIFF
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -1,5 +1,6 @@
 package com.gu.ssm
 
+import com.amazonaws.regions.Region
 import com.gu.ssm.utils.attempt._
 
 import scala.concurrent.duration._
@@ -24,7 +25,7 @@ object Main {
               case _ => fail
             }
           case SsmSsh => bastionInstanceIdOpt match {
-            case None => setUpStandardSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent)
+            case None => setUpStandardSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent, profile, region)
             case Some(bastionInstance) => setUpBastionSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, bastionInstance, bastionPortNumberOpt, bastionUser, targetInstancePortNumberOpt, useAgent, preferredAlgs)
           }
           case SsmScp => (sourceFileOpt, targetFileOpt) match {
@@ -55,7 +56,9 @@ object Main {
     rawOutput: Boolean,
     targetInstancePortNumberOpt: Option[Int],
     preferredAlgs: List[String],
-    useAgent: Option[Boolean]) = {
+    useAgent: Option[Boolean],
+    profile: Option[String],
+    region: Region) = {
     val fProgramResult = for {
       config <- IO.getSSMConfig(awsClients.ec2Client, awsClients.stsClient, executionTarget)
       sshArtifacts <- Attempt.fromEither(SSH.createKey())
@@ -70,7 +73,7 @@ object Main {
       address <- Attempt.fromEither(Logic.getAddress(instance, onlyUsePrivateIP))
       hostKeyFile <- SSH.writeHostKey((address, hostKey))
     } yield {
-      SSH.sshCmdStandard(rawOutput)(privateKeyFile, instance, user, address, targetInstancePortNumberOpt, Some(hostKeyFile), useAgent)
+      SSH.sshCmdStandard(rawOutput)(privateKeyFile, instance, user, address, targetInstancePortNumberOpt, Some(hostKeyFile), useAgent, profile, region)
     }
     val programResult = Await.result(fProgramResult.asFuture, maximumWaitTime)
     ProgramResult(programResult.map(UI.sshOutput(rawOutput)))
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -13,7 +13,7 @@ object Main {
 
   def main(args: Array[String]): Unit = {
     val (result, verbose) = argParser.parse(args, Arguments.empty()) match {
-      case Some(Arguments(verbose, Some(executionTarget), toExecuteOpt, profile, region, Some(mode), Some(user), sism, _, _, onlyUsePrivateIP, rawOutput, bastionInstanceIdOpt, bastionPortNumberOpt, Some(bastionUser), targetInstancePortNumberOpt, useAgent, preferredAlgs, sourceFileOpt, targetFileOpt)) =>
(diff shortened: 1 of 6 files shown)


shtukas:
Nice one!

katebee:
Hello 👋 are we happy to get this merged and released?

mbarton:
Yep definitely! I'm not sure about the user friendliness of the extra parameter but we (in the Investigations team) can use wrapper scripts in the meantime.

I will find some time to re-test and release in the next couple of weeks

adamnfish:
I think this wants a README update as well. At least an update to the dumped "usage" section, but it would be good to document the flag. There's also a big section no "bastions" and I think this change has some relevance to the content in there.

stiliyana-simeonova:
Just to say that with me it failed to compile without a minor change in Logic.scala (using .linesIterator instead of .lines)

OS: Debian 10 (buster)
JDK: openjdk version "11.0.6" 2020-01-14

mbarton:
Funky, it compiles for me. We're you compiling using SBT @stiliyana-simeonova? I wonder if you've compiled against Scala 2.13 rather than 2.12

mbarton:
All tested and works in the investigations internet-enabled VPC. The [latest Ubuntu base images](https://cloud-images.ubuntu.com/locator/) also seem to have an up to date version with support so I will update the [base image](https://amigo.gutools.co.uk/base-images/ubuntu-bionic%20(18.04%20LTS))

mbarton:
Ah @stiliyana-simeonova just saw that a test is failing on my branch which might have been the error you saw?

It's a test for a helper function so I'm not sure what I've done to perturb it. Will take a look

stiliyana-simeonova:
> Funky, it compiles for me. We're you compiling using SBT @stiliyana-simeonova? I wonder if you've compiled against Scala 2.13 rather than 2.12

That is indeed the case :)
Thanks

stiliyana-simeonova:
> Ah @stiliyana-simeonova just saw that a test is failing on my branch which might have been the error you saw?
> 
> It's a test for a helper function so I'm not sure what I've done to perturb it. Will take a look

it was compile failure rather than test failure. I will try compiling against 12.2 and feed back in a bit

mbarton:
The test seems flakey, mostly passing and sometimes failing on both master and my branch. I've tried changing the timeout but that doesn't seem to fix it.

I think I'm going to merge and release this anyway

mbarton:
Released [v1.6.0](https://github.com/guardian/ssm-scala/releases/tag/v1.6.0) with this change. PR to update our brew formula: https://github.com/guardian/homebrew-devtools/pull/46

Thanks to the team for the clear release instructions 💯

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -24,11 +25,11 @@ object Main {
...
-            case None => setUpStandardSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent)
+            case None => setUpStandardSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent, profile, region, tunnelThroughSystemsManager)
             case Some(bastionInstance) => setUpBastionSSH(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, bastionInstance, bastionPortNumberOpt, bastionUser, targetInstancePortNumberOpt, useAgent, preferredAlgs)
           }
           case SsmScp => (sourceFileOpt, targetFileOpt) match {
-            case (Some(sourceFile), Some(targetFile)) => setUpStandardScp(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent, sourceFile, targetFile)
+            case (Some(sourceFile), Some(targetFile)) => setUpStandardScp(awsClients, executionTarget, user, sism, onlyUsePrivateIP, rawOutput, targetInstancePortNumberOpt, preferredAlgs, useAgent, sourceFile, targetFile, profile, region, tunnelThroughSystemsManager)
```
rtyley: I'm wondering if this means that ssm-tunneling is supported here for SCP from an end-user perspective? Like, it seems `ssm` can use ssm-tunneling to install the SSH key - but then when executing the `scp` step, access won't be over the ssm-tunnel will it?

mbarton: Ah whoops, I haven't added it to the argument parser for any commands other than `ssh`. Will raise a follow up PR.

With that fixed, it should work as we set up the proxy command for scp [here](https://github.com/guardian/ssm-scala/blob/master/src/main/scala/com/gu/ssm/SSH.scala#L161). Provisioning the key is done via a separate [ssm run-command](https://github.com/guardian/ssm-scala/blob/master/src/main/scala/com/gu/ssm/Main.scala#L141) API call so won't be tunnelled through the ssh session eventually established for the user.

mbarton: https://github.com/guardian/ssm-scala/pull/113

rtyley: That sounds awesome I look forward to trying it out tomorrow!

mbarton: Release v1.6.1 with this in @rtyley. Give it a spin:

```
brew upgrade guardian/devtools/ssm

# Example to download a file from remote machine:
ssm scp -x -i <instance_id> --ssm-tunnel --profile investigations :mtest.txt ./mtest.txt
```

PR #116, opened 2020-05-11
Add --ssm-tunnel to quick start docs

philmcmahon:
Wherever possible teams should be using --ssm-tunnel rather than IP restrictions to control instance access so maybe it deserves a place in the quick start docs? 

Counter argument: let's leave this out and wait till *everyone* is using --ssm-tunnel and then make that the default behaviour

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,13 @@ The readme is quite detailled (and shows how to do many more things than what wi
 3. At your console type
 
 	```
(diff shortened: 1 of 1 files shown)


shtukas:
Um... Interesting one.

shtukas:
I vote for adding it.

philmcmahon:
Thanks for the review @shtukas , I've got another pr coming soon to replace this!

PR #120, opened 2020-08-28
Allow user to just provide one tag rather than all 3.

philmcmahon:
## What does this change?
This change follows from https://github.com/guardian/ssm-scala/pull/117 in an eternal quest to make ssh a bit easier. My proposal here is that we allow people to only specify one tag. For example it is now possible to do:

`ssm ssh -x -p deployTools -t grafana`
`ssm ssh -x -p deployTools -t prism,prod --newest`
`ssm ssh -x -p deployTools -t prod,prism --newest`
`ssm ssh -x -p deployTools -t api,code,flexible --newest`

I've also made the describeinstances call even more inefficient by trying to make it less case sensitive, so

`ssm ssh -x -p deployTools -t prism,prod --newest` will actually query:

describeinstances( app=prism|PRISM|prod|PROD, stage=prism|PRISM|prod|PROD)

I've tried to keep order as irrelevant as possible, but without doing multiple API lookups I needed a way to decide which filter to not include if less than 3 tags are provided. So if only one tag is provided, only the App tag is checked. if two are provided, then App and stage are checked (-t grafana,prod and -t prod,grafana are equivalent). If 3 tags are provided then the order doesn't matter.

## What is the value of this?
Faster SSH. This is partly just a 'nice to have' feature to encourage people to switch to SSM tunnel, which will become the default in the next release.

### DIFF
--- a/src/main/scala/com/gu/ssm/Logic.scala
+++ b/src/main/scala/com/gu/ssm/Logic.scala
@@ -20,8 +20,8 @@ object Logic {
   }
 
   def extractSASTags(tags: Seq[String]): Either[String, List[String]] = {
-    if (tags.length != 3) Left("Please provide app, stack and stage tags")
-    else if (tags.head.length == 0) Left("Please supply an app, stack and stage tag in any order. For example, -t grafana,PROD,deploy")
+    if (tags.length > 3 || tags.isEmpty || tags.head.length == 0) Left("Please supply at least one and no more than 3 tags. " +
+      "If you specify less than 3 tags order assumed is app,stage,stack")
     else Right(tags.toList)
   }
 
--- a/src/main/scala/com/gu/ssm/aws/EC2.scala
+++ b/src/main/scala/com/gu/ssm/aws/EC2.scala
@@ -32,7 +32,7 @@ object EC2 {
     val allTags = tagValues ++ tagValues.map(_.toUpperCase) ++ tagValues.map(_.toLowerCase)
 
     // if user has provided fewer than 3 tags then assume order app,stage,stack
-    val tagOrder = List("App", "Stack", "Stage")
+    val tagOrder = List("App", "Stage", "Stack")
     val filters = tagOrder.take(tagValues.length).map(makeFilter(_, allTags))
 
     val request = new DescribeInstancesRequest()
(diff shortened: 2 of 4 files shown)


src/main/scala/com/gu/ssm/aws/EC2.scala
```diff
@@ -26,13 +26,18 @@ object EC2 {
       .build()
   }
 
+  def makeFilter(tagName: String, values: List[String]) = new Filter(s"tag:$tagName", values.asJava)
+
   def resolveByTags(tagValues: List[String], client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[List[Instance]] = {
+    val allTags = tagValues ++ tagValues.map(_.toUpperCase) ++ tagValues.map(_.toLowerCase)
+
+    // if user has provided fewer than 3 tags then assume order app,stage,stack
```
akash1810, line 34: The comment doesn't match the tag order below.

```suggestion
    // if user has provided fewer than 3 tags then assume order app,stack,stage
```

src/main/scala/com/gu/ssm/Logic.scala
```diff
@@ -20,8 +20,8 @@ object Logic {
   }
 
   def extractSASTags(tags: Seq[String]): Either[String, List[String]] = {
-    if (tags.length != 3) Left("Please provide app, stack and stage tags")
-    else if (tags.head.length == 0) Left("Please supply an app, stack and stage tag in any order. For example, -t grafana,PROD,deploy")
+    if (tags.length > 3 || tags.isEmpty || tags.head.length == 0) Left("Please supply at least one and no more than 3 tags. " +
```
akash1810, line 23: I might be missing something, but Is `tags.isEmpty` and `tags.head.length == 0` the same thing?

akash1810 (approved):
Couple of minor comments, LGTM otherwise. Excited!

src/main/scala/com/gu/ssm/Logic.scala
philmcmahon: The idea is to check for List("") as well as List(). I think the argument lib would never allow List("") to happen but this is just belt and braces sort of thing

src/main/scala/com/gu/ssm/aws/EC2.scala
philmcmahon: ah good spot, thanks! This was actually the order I wanted so I've corrected the line below.

philmcmahon:
Depdendent on https://github.com/guardian/ssm-scala/pull/121

PR #122, opened 2020-08-28
Add github actions workflow to compile and test

philmcmahon:
## What does this change?
Adds a workflow to run `sbt test`. 

## What is the value of this?
This should reduce the risk of merging broken code as we'll get a red cross on the PR (to go with the semi permanent snyk ❌ ...)

### DIFF
--- a/.github/workflows/compile-test.yml
+++ b/.github/workflows/compile-test.yml
@@ -4,16 +4,11 @@ on:
   workflow_dispatch:
   push:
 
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
Looks like there's a first-party action too: https://github.com/actions/starter-workflows/blob/8e9dd416eb038512da8d6571d57d69d1d73c2ff8/ci/scala.yml

PR #123, opened 2020-08-28
Make --ssm-tunnel default behaviour.

philmcmahon:
## What does this change?
This modifies ssm-scala so that by default it tries to proxy requests via aws systems manager. This is the preferred approach, as documented in the [readme](https://github.com/guardian/ssm-scala#enabling-ssm-tunnel) as it means you don't have to open up port 22, don't need the VPN, and don't need bastion hosts. Oh my! 

I think enough time has passed since @mbarton 's initial PR https://github.com/guardian/ssm-scala/pull/111 for this to have become mostly standard practice, and this PR will save people having to remember to type `--ssm-tunnel` all the time. 

If ssm tunnel is not desired, the flag `--no-ssm-proxy` will disable this behaviour.

This combined with the work discussed here https://github.com/guardian/ssm-scala/pull/120 will allow you to ssh like this:

`ssm ssh -x -t grafana -p deployTools` 

## What is the value of this?
Firstly, it saves typing a few characters. Secondly, it pushes people towards a more convenient and secure way of accessing machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Usage: ssm [cmd|repl|ssh|scp] [options] <args>...
 
   -p, --profile <value>    The AWS profile name to use for authenticating this execution
   -i, --instances <value>  Specify the instance ID(s) on which the specified command(s) should execute
-  -t, --tags <value>       Search for instances by tag e.g. '--tags app,stack,stage'
+  -t, --tags <value>       Search for instances by tag e.g. '--tags app,stage,stack'
   -r, --region <value>     AWS region name (defaults to eu-west-1)
   --verbose                enable more verbose logging
 Command: cmd [options]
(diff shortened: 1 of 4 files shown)


akash1810 (approved):

README.md
```diff
@@ -85,14 +85,14 @@ Create and upload a temporary ssh key
...
   -B, --bastion-tags <value>
-                           Connect through the given bastion identified by its tags; implies -a (use agent) unless followed by -A. --ssm-tunnel can be used to avoid the need for a bastion instance
-  --bastion-port <value>   Connect through the given bastion at a given port. --ssm-tunnel can be used to avoid the need for a bastion instance
-  --bastion-user <value>   Connect to bastion as this user (default: ubuntu). --ssm-tunnel can be used to avoid the need for a bastion instance
+                           Connect through the given bastion identified by its tags; implies -a (use agent) unless followed by -A.
+  --bastion-port <value>   Connect through the given bastion at a given port.
+  --bastion-user <value>   Connect to bastion as this user (default: ubuntu).
   --host-key-alg-preference <value>
                            The preferred host key algorithms, can be specified multiple times - last is preferred (default: ecdsa-sha2-nistp256, ssh-rsa)
-  --ssm-tunnel             Connect to the host proxying through AWS Systems Manager, rather than directly to port 22. Requires Systems Manager Agent > 2.3.672.0 to be installed.
```
mbarton, line 95: Is it worth keeping the argument (undocumented) to avoid breaking scripts that rely on it?

To be fair, there's only 9 hits for it in our GitHub organisation: https://github.com/search?q=org%3Aguardian+%22--ssm-tunnel%22&type=Code

mbarton (commented):
I'm onboard with this 👍 

It may not be much more effort to have ssm "just work" by reading a value from Prism that says whether the instance can support tunnelling (Agent Version is new enough, correct IAM policies applied). It could fall back to the old route if not and log a warning to the user linking to some docs to upgrade.

Would be quite swag but I'll leave it up to you to decide whether it's worth it or not. Certainly a way of surveying our estate to see how far the SSM tunnel roll out has got is worth it - but maybe we've already done that by hand or asking teams individually?

README.md
philmcmahon: good idea

Nirvikalpa108: Thank you so much for this great comment @mbarton! We've kept the `--ssm-tunnel` argument in place, as you advised, with no implementation so that it doesn't break any scripts reliant upon it. Great spot, thank you! 👍🏽

We investigated checking whether the instance has ssm permissions, but we decided not to go ahead and implement this for a couple of reasons. We didn't want to add prism as a dependency as that would require vpn access. 

Second, we checked whether we could use the AWS SDK, but this requires us to make 4 separate API calls (`describe-instances`, `get-profile`, `get-role` and `get-policy`) and we felt this was perhaps too much complexity for this feature. 

Thank you ❤️

PR #125, opened 2021-03-25
Wait indefinitely

mbarton:
In the investigations account we regularly hit this timeout simply running `ssm ssh`. I think this is because of the time it takes to establish a connection to the SSM agent as subsequent invocations do not time out. When the timeout does occur, ssm hangs rather than exits.

We'd also like to run scripts using `ssm cmd` that will take longer than 25 seconds.

I don't see value in having an arbitrary timeout across the entire program so after this change ssm will wait forever. I may well be missing the reason behind this timeout though.


shtukas (approved):

mbarton:
I don't have access to look and see what snyk is complaining about here unfortunately. I assume it's an existing issue

katebee (approved):
👍  I am also a fan of removing the timeout

PR #127, opened 2021-04-06
Return error code if any invocation fails during ssm cmd

mbarton:
At the moment, `ssm cmd` will return exit code 0 even if the underlying command has failed.

I'm trying to use it to run a script on a brand new EC2 instance, where the script is downloaded and installed by the launch configuration. If `ssm cmd` runs before the script has been downloaded, the command will result in an error (verified in the SSM web UI) but `ssm cmd` returns a success code.

It's also common to `set -e` in bash scripts to abandon execution on the first failure so the current behaviour can lead to cascading failures in subsequent commands.

This change returns exit code 1 if the command has failed on any of the targets. The stdout returned from the SSM agent will indicate the reason for the failure.

### DIFF
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -76,7 +76,7 @@ object Main {
       SSH.sshCmdStandard(rawOutput)(privateKeyFile, instance, user, address, targetInstancePortNumberOpt, Some(hostKeyFile), useAgent, profile, region, tunnelThroughSystemsManager)
     }
     val programResult = Await.result(fProgramResult.asFuture, Duration.Inf)
-    ProgramResult(programResult.map(UI.sshOutput(rawOutput)))
+    ProgramResult.convertErrorToResult(programResult.map(UI.sshOutput(rawOutput)))
   }
 
   private def setUpBastionSSH(
@@ -115,7 +115,7 @@ object Main {
       hostKeyFile <- SSH.writeHostKey((bastionAddress, bastionHostKey), (targetAddress, targetHostKey))
     } yield SSH.sshCmdBastion(rawOutput)(privateKeyFile, bastionInstance, targetInstance, user, bastionAddress, targetAddress, bastionPortNumberOpt, bastionUser, targetInstancePortNumberOpt, useAgent, Some(hostKeyFile))
     val programResult = Await.result(fProgramResult.asFuture, Duration.Inf)
-    ProgramResult(programResult.map(UI.sshOutput(rawOutput)))
+    ProgramResult.convertErrorToResult(programResult.map(UI.sshOutput(rawOutput)))
   }
 
   private def setUpStandardScp(
@@ -150,7 +150,7 @@ object Main {
       SSH.scpCmdStandard(rawOutput)(privateKeyFile, instance, user, address, targetInstancePortNumberOpt, useAgent, Some(hostKeyFile), sourceFile, targetFile, profile, region, tunnelThroughSystemsManager)
     }
     val programResult = Await.result(fProgramResult.asFuture, Duration.Inf)
-    ProgramResult(programResult.map(UI.sshOutput(rawOutput)))
+    ProgramResult.convertErrorToResult(programResult.map(UI.sshOutput(rawOutput)))
   }
 
   private def execute(awsClients: AWSClients, executionTarget: ExecutionTarget, user: String, toExecute: String): ProgramResult = {
@@ -159,10 +159,9 @@ object Main {
       _ <- Attempt.fromEither(Logic.checkInstancesList(config))
       results <- IO.executeOnInstances(config.targets.map(i => i.id), user, toExecute, awsClients.ssmClient)
       incorrectInstancesFromInstancesTag = Logic.computeIncorrectInstances(executionTarget, results.map(_._1))
-      hasAnyCommandFailed = Logic.hasAnyCommandFailed(results)
-    } yield ResultsWithInstancesNotFound(results, incorrectInstancesFromInstancesTag, hasAnyCommandFailed)
+    } yield ResultsWithInstancesNotFound(results, incorrectInstancesFromInstancesTag)
 
     val programResult = Await.result(fProgramResult.asFuture, Duration.Inf)
-    ProgramResult(programResult.map(UI.output))
+    ProgramResult.convertErrorToResult(programResult.map(UI.output))
   }
 }
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -58,7 +58,7 @@ object Main {
     useAgent: Option[Boolean],
     profile: Option[String],
     region: Region,
-    tunnelThroughSystemsManager: Boolean) = {
+    tunnelThroughSystemsManager: Boolean): ProgramResult = {
     val fProgramResult = for {
       config <- IO.getSSMConfig(awsClients.ec2Client, awsClients.stsClient, executionTarget)
       sshArtifacts <- Attempt.fromEither(SSH.createKey())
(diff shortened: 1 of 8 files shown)


shtukas (approved):

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -160,7 +160,13 @@ object Main {
       results <- IO.executeOnInstances(config.targets.map(i => i.id), user, toExecute, awsClients.ssmClient)
       incorrectInstancesFromInstancesTag = Logic.computeIncorrectInstances(executionTarget, results.map(_._1))
     } yield ResultsWithInstancesNotFound(results,incorrectInstancesFromInstancesTag)
+
     val programResult = Await.result(fProgramResult.asFuture, Duration.Inf)
-    ProgramResult(programResult.map(UI.output))
+    val programOutput = ProgramResult(programResult.map(UI.output))
+
+    val anyCommandFailed = programResult.exists(_.results.exists(_._2.map(_.commandFailed).getOrElse(false)))
```
nicl, line 167: Minor, but this kind of Scala line can be pretty impenetrable (nested calls, `_`, and `_2` for the tuple).

adamnfish, line 167: I think this logic should go in the for loop as part of the program, rather than being tacked onto the end after we've awaited the Future.

Eliminating the effect (Future in this case) should be as close as possible to the edge of your program - the last thing you do. Sticking to this approach naturally forces us to write programs that can be easily tested and reasoned about, because it encourages us to separate our logic form our side-effects. In this case, the fiddly logic that Nic mentions above can't be tested, because it is entwined with the effectful code at the edge of our program.

Would it be neater if we move this logic to a function and call that function as another step in our program's execution? The for comprehension lists all the steps our program needs to perform, and returns a datastructure that describes the result of the program's execution. We'll likely have to use a different (or changed) datastructure to include this failure representation so it's a bit more work than simply tacking the logic onto the end, but I think it will make this program more reliable and maintainable for the future.

mbarton: I totally agree and I've modified it to be a separate tested function in `Logic.scala`. It still has nested `.exists` though which might still be to brief for it's own good. What do you think?

mbarton: I agree. I've refactor the logic to a helper function and each for comprehsion now returns a `ProgramResult`. Let me know what you think

src/main/scala/com/gu/ssm/UI.scala
```diff
@@ -60,16 +59,18 @@ object UI {
         )
       }
     }
-    buffer.toList
+
+    val nonZeroExitCode = if(extendedResults.anyCommandFailed) { Some(ErrorCode) } else { None }
```
adamnfish, line 63: #ScalaStyle no need for braces on a single line like this, and a space between the `if` and its parens would be more normal.

```scala
val nonZeroExitCode = if (extendedResults.anyCommandFailed) Some(ErrorCode) else None
```

If it were long enough (I'm not saying it is) I'd use two lines:
```scala
val nonZeroExitCode =
  if (extendedResults.anyCommandFailed) Some(ErrorCode)
  else None
```

No harm, though!

src/main/scala/com/gu/ssm/models.scala
```diff
@@ -84,7 +84,7 @@ case object SsmRepl extends SsmMode
 case object SsmSsh extends SsmMode
 case object SsmScp extends SsmMode
 
-case class CommandResult(stdOut: String, stdErr: String)
+case class CommandResult(stdOut: String, stdErr: String, commandFailed: Boolean)
```
adamnfish, line 87: Is it a bit "double negative" to store this as `commandFailed`? For most of the usages I don't min in the slightest but there are a couple of examples in the test data that made me squint because false = ✅  and true = 🚫 .

I haven't checked the code out and tried both ways so whatever you think is fine. If this way round makes it easier where it's actually used then that makes great sense.

src/main/scala/com/gu/ssm/UI.scala
```diff
@@ -60,16 +59,18 @@ object UI {
...
+    ProgramResult(buffer.toList, nonZeroExitCode)
   }
 
-  def sshOutput(rawOutput: Boolean)(result: (InstanceId, Seq[Output])): Seq[Output] = {
+  def sshOutput(rawOutput: Boolean)(result: (InstanceId, Seq[Output])): ProgramResult = ProgramResult(
```
adamnfish, line 67: This signature change is surprising to me, because the places it is called from are already wrapping the result in `ProgramResult`. 
e.g. https://github.com/guardian/ssm-scala/pull/127/files#diff-5f0d672d660b66a9766e0cffc1f5888e92888015b8198a7369a7dbc5520adb0aR153

I think this is a sign that the existing `ProgramResult` class wants some thought, because it seems like the codebase kind of flips between that and `Seq[Output]`, with the two being semi-interchangeable. It looks like `ProgramResult` might really be a function (its `apply`) disguised as a datastructure...

If this stems from existing confusion then no worries at all and we can take a look separately, as long as you're happy that introducing multi-nesting of the `ProgramResult` isn't making things more complicated here, now.

src/main/scala/com/gu/ssm/models.scala
```diff
@@ -97,7 +97,11 @@ case class AWSClients (
   ec2Client: AmazonEC2Async
 )
 
-case class ResultsWithInstancesNotFound(results: List[(InstanceId, scala.Either[CommandStatus, CommandResult])], instancesNotFound: List[InstanceId])
+case class ResultsWithInstancesNotFound(
+  results: List[(InstanceId, scala.Either[CommandStatus, CommandResult])],
+  instancesNotFound: List[InstanceId],
+  anyCommandFailed: Boolean
```
adamnfish, line 103: I always sit up when I see a field added that duplicates existing information. "could this be a function instead"?

In this case I think the answer is yes, and that function has been added in this PR (`hasAnyCommandFailed`).

Would it be simpler to use the existing datastructure and call `hasAnyCommandFailed` from the UI? The if statement that follows is checking the `anyCommandFailed` property, but it could instead check `hasAnyCommandFailed(extendedResults.results)`.
https://github.com/guardian/ssm-scala/pull/127/files#diff-b05affdbb83b2b17713b012ad87e1df83566bf9690471f545f97ccfde4945095R63

Trade-offs! It's nice to have the logic in the program and the UI do minimal work, which is what you've done here. Great stuff!
When this means duplicating info is when it's not clear cut. One downside to consider about duplication in datastructures is that this then needs to be kept in sync. If someone creates a copy of the `ResultsWithInstancesNotFound` to change the results they'd need to remember to update the `anyCommandFailed` member (and vice-versa).

src/test/scala/com/gu/ssm/LogicTest.scala
```diff
@@ -158,4 +158,23 @@ class LogicTest extends FreeSpec with Matchers with EitherValues {
       }
     }
   }
+
+  "hasAnyCommandFailed" - {
```
adamnfish, line 162: Great!

src/main/scala/com/gu/ssm/models.scala
mbarton: I think it's the best way round to avoid having to negate the `exists` check in `hasAnyCommandFailed`:

```scala
ssmResults.exists { case(_, result) => result.exists(_.commandFailed) }
```

src/main/scala/com/gu/ssm/UI.scala
mbarton: Yeah I started returning `ProgramResult` directly as it already had a field for the exit code. As you say the `apply` method that turns the `Left[FailedAttempt, _]` into a `ProgramResult` isn't really in the spirit of an `apply` method any more and so it looks like double wrapping in the caling code. I've renamed the method to `convertErrorToResult` to hopefully make it clearer what it is doing after this change

src/main/scala/com/gu/ssm/models.scala
mbarton: Yep, great shout. Calling the function on-demand also means I don't need to modify the interactive UI code that doesn't care about an exit code at all

mbarton: I also moved the function to `UI` from `Logic` as it's only called within `UI`

adamnfish (approved):
Thank you!

PR #129, opened 2021-04-13
Use default AWS credentials if no profile is set

mbarton:
We are automating AMI rotation for our neo4j instance in Giant using a bash script that calls `ssm cmd` to run commands. To run this automatically I'd like to simply run the same script in AWS but unfortunately I get an error:

```
--profile switch or environment variable AWS_PROFILE expected
```

It doesn't help if I `export AWS_PROFILE=default` as AWS credentials are not provided using a file but via the EC2/ECS metadata service:

```
"The config profile (default) could not be found"
```

This PR adds a `--use-default-credentials-provider` flag which we can set if running the script in AWS itself. We've deliberately decided to deviate from the standard `aws` CLI behaviour here so that people can't unexpectedly run commands using credentials in the default profile if they forget the `--profile` argument.

### DIFF
--- a/src/main/scala/com/gu/ssm/aws/AwsAsyncHandler.scala
+++ b/src/main/scala/com/gu/ssm/aws/AwsAsyncHandler.scala
@@ -28,7 +28,7 @@ object AwsAsyncHandler {
       if (e.getMessage.contains("Request has expired")) {
         Failure("expired AWS credentials", "Failed to request data from AWS, the temporary credentials have expired", AwsPermissionsError, e).attempt
       } else if (e.getMessage.contains("Unable to load AWS credentials from any provider in the chain")) {
-        Failure("Invalid AWS profile name (no credentials)", "No credentials found for the specified AWS profile", AwsPermissionsError, e).attempt
+        Failure("No AWS credentials found", "No AWS credentials found. Did you mean to set --profile?", AwsPermissionsError, e).attempt
       } else if (e.getMessage.contains("No AWS profile named")) {
         Failure("Invalid AWS profile name (does not exist)", "The specified AWS profile does not exist", AwsPermissionsError, e).attempt
       } else if (e.getMessage.contains("is not authorized to perform")) {
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -51,6 +51,11 @@ object ArgumentParser {
     opt[Unit]("verbose").action( (_, c) =>
       c.copy(verbose = true) ).text("enable more verbose logging")
 
+    opt[Unit]("use-default-credentials-provider").optional()
+      .action((value, args) => args.copy(useDefaultCredentialsProvider = true))
+      .text("Use the default AWS credentials provider chain rather than profile credentials. " +
+            "This option is required when running within AWS itself.")
+
     cmd("cmd")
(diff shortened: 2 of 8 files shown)


shtukas:
+1

shtukas (approved):

adamnfish (commented):
I'd much prefer an alternate flag that instructs `ssm` to use the default provider chain. One of the nice things about us always using named profiles is that it prevents errors from people accidentally using a default profile. I can see why forcing the default chain would be useful, but I think that should be an explicit choice by the user.

mbarton:
As discussed in chat, this PR changes behaviour to match the `aws` cli which is probably less surprising and in the error message it mentions in the `--profile` option needed locally.

@adamnfish you mentioned on chat there were some accidents that could happen with the use of the default profile locally. Could you expand a bit more on what those are? As Janus gives out credentials with a named profile by default Guardian users would only have credentials in the default profile if they had done so deliberately.

adamnfish:
I often see people (including myself) accidentally omit the `profile` parameter when running AWS commands. For the most part we dodge the problem because as you say, Janus only provides named profiles by default. However, it's common for people to have a default profile, even so. Perhaps they followed a tutorial on the internet while working against the dev playground. Maybe they are using their own account for a side project. I know in the past some teams have encouraged colleagues to use the default profile to save time.

I think the default profile name is a trap! I understand why AWS need to provide it, because for the numerical majority of their users that's all people need. For the users that make up the bulk of AWS' activity (like us!), the default profile should be avoided. Since `ssm` is aimed at the latter group, I think it should prioritise safety and require explicit action to use the default chain.

I'd observe that explicit use of the default profile would serve the described use-case perfectly well. If you're scripting this activity then it's no harm to script the explicit requirement to do something unusual / less safe.

I don't feel as strongly as I sound here though, and I appreciate that consistency with the AWS CLI is a virtue! Happy either way, perhaps we could chat about it over a call if we aren't sure?

mbarton:
OK, that's not a problem I've come across myself but I wouldn't want anyone accidentally running something destructive against the wrong account. I've updated the PR to control it using a `--use-default-credentials-provider` flag.

Could you have a quick look @adamnfish?

src/main/scala/com/gu/ssm/aws/AwsAsyncHandler.scala
```diff
@@ -28,7 +28,7 @@ object AwsAsyncHandler {
       if (e.getMessage.contains("Request has expired")) {
         Failure("expired AWS credentials", "Failed to request data from AWS, the temporary credentials have expired", AwsPermissionsError, e).attempt
       } else if (e.getMessage.contains("Unable to load AWS credentials from any provider in the chain")) {
-        Failure("Invalid AWS profile name (no credentials)", "No credentials found for the specified AWS profile", AwsPermissionsError, e).attempt
+        Failure("No AWS credentials found", "No AWS credentials found. Did you mean to set --profile?", AwsPermissionsError, e).attempt
```
adamnfish, line 31: Do we want to say something like "or specify `--use-default-credentials-provider`" here as well? Minor point and no stressin'!

adamnfish (approved):
❤️  🙏

src/main/scala/com/gu/ssm/aws/AwsAsyncHandler.scala
mbarton: Yeah I thought about it but decided against it because it's a niche use case that people won't want to use when running locally

PR #132, opened 2021-04-29
correct the permissions needed in the readme

johnduffell:
This PR adds some more perms to the readme, as it turns out they are needed to avoid a future timing out after 25 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The readme is quite detailed (and shows how to do many more things than what wil
 
 ## Known issues
 
-If you get an error about Futures timed out after 25 seconds, then the SSM permissions may not be right, or you might need to recycle the instance since adding them.
+If you get an error about Futures timed out after 25 seconds, then the SSM permissions may not be right, or you might need to recycle the instance since adding the permissions.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,10 +281,10 @@ You'll also need to install the systems manager plugin on your machine:
   brew cask install session-manager-plugin
 ```
 
-You can then SSH using SSM and the --ssm-tunnel command:
+You can then SSH using SSM with the default arguments:
```
adamnfish, line 284: Great catch! This changed to the default a while back, thanks.

README.md
```diff
@@ -51,6 +51,8 @@ The readme is quite detailed (and shows how to do many more things than what wil
 
 ## Known issues
 
+If you get an error about Futures timed out after 25 seconds, then the SSM permissions may not be right, or you might need to recycle the instance since adding them.
```
adamnfish, line 54: Might help to clarify the word `them` at the end here, to make it clear that we're talking about the changed permissions taking effect? Minor point though and this is fine!

adamnfish (approved):

PR #137, opened 2021-12-16
chore: Bump dependencies to latest minor version available

akash1810:
## What does this change?

Bumps dependencies to latest minor version available.

Note:
  - It looks like `scalatest` v3.2.9 has a different package layout, so imports across the tests have also been updated.
  - `scopt` has a v4 release. As this is a major change it's likely to include breaking changes and should be upgraded separately.


nicl (approved):
Thanks for doing this!

PR #245, opened 2023-03-17
Add commands for SSH tunnelling

jonathonherbert:
## What does this change?

Adds two options to enable SSH tunnelling. They're mutually exclusive.

The first, `--tunnel localPort:hostname:remotePort`, allows the user to specify an arbitrary hostname and remote port.
The second, `--rds-tunnel localPort:list,of,tags`, allows the user to specify a local port, and a list of tags. SSM looks up an RDS instance via the tags, failing if there are 0 or many instances.

## What is the value of this?

Across our estate, we have many scripts that accomplish this in a few different ways – here are [a few examples,](https://github.com/search?q=org%3Aguardian+path%3Assh-tunnel&type=code) and I suspect there are more. These commands aim to replace most of them with a one-liner that can be included in project README files and used directly.

## Any additional notes?

There are a few decisions here that gave me pause. One concerns the interface.

~When using -x to run the command immediately, there's no output to affirm that the **correct hostname has been chosen** (important with a tag lookup, or to aid debugging of any sort), or to e.g. **warn the user when connecting to resources with a `PROD` tag.** I've spent some time looking but do not yet understand how this program pipes output to the shell – to me, both these features are dealbreakers, and I'd be delighted to pair with somebody who did know how we might accomplish this.~

This is now added. (To cover the unknown above, the script adds its output by piping directly to the shell via a wrapper script – see https://github.com/guardian/ssm-scala/pull/61.) Example output tunnelling to a DB with PROD tags:

```
# If the command succeeded, a tunnel has been established.
# Local port: 5432
# Remote address: typerighter-rule-manager-store-prod.cbyyqud0gzor.eu-west-1.rds.amazonaws.com:5432
# The tags indicate that this is a PRODUCTION resource. Please take care! Perhaps bring a pair?
```

Unlike other ssh commands, this output does still appear when `--raw` is specified. As it goes to stderr via `Metadata`, it doesn't interfere when parsed out with `eval`, so I hope that's fine.

There were also a few modelling decisions:

- I went between a few options in modelling these values, and settled on two case classes, `TunnelTargetWithRDSTags` and  `TunnelTargetWithHostName`, where the former resolves into the latter as the program executes.
- We carry through two optional parameters in the `Arguments` case class, rather than one. I felt it was neater with one, but having two allows us to fail when both options are specified via `checkConfig`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,26 @@ ssm ssh --profile security -t security-hq,security,PROD --newest --execute
 
 instead of the example given in the previous `--raw` section.
 
+### --tunnel
+
+This flag forwards traffic from a local port through the instance to the specified hostname and port. For example,
+
+```
+ssm ssh --profile security -t security-hq,security,PROD --newest -x --tunnel 5000:example.com:6000
+```
+
+would forward all traffic on your machine through the remote instance to example.com:6000.
+
+### ---rds-tunnel
+
+Similar to `tunnel`, this flag forwards traffic from a local port to an AWS RDS database specified by the given tags. For example,
+
+```
+ssm ssh --profile security -t security-hq,security,PROD --newest -x --rds-tunnel 5000:example-db,security,CODE
+```
+
+would try to find a single RDS instance with the tags `example-db,security,CODE`, and forward traffic from port 5000 to that RDS instance via the remote instance.
+
 ## Disabling SSM Tunnel
 **By default, SSM proxies your connection via AWS systems manager**, which saves you from opening up port 22, connecting to
 the VPN, or using bastion hosts. This requires a recent version of systems manager to be runnning on your machine and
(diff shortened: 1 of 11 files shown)


philmcmahon (commented):
Just to say I've looked through the scala here and would be happy with merging this as it would be useful right now. Understand it might be worth coming up with a helpful output solution though

shtukas (approved):
Pragmatic solution to a real problem. Essentially similar comment as Phil, but happy to approve. Nice work!

kenoir (commented):
@jonathonherbert I think you're ok to merge this and iterate.

jonathonherbert:
Thanks all. It wasn't too far out of my way, so 9595a08f256dee25fe9884b2de5bb7fa156a34d7 adds the additional output, now described in the PR.

jonathonherbert:
I'll wait for another tick before I merge, as another pair of 👀 on the messaging I've added seems sensible.

shtukas (approved):

PR #375, opened 2024-01-05
Make -x (execute) the default behaviour

philmcmahon:
## What does this change?
This PR modifies ssm-scala so that you no longer need to pass `-x` or `--execute` in order to get it to execute the ssh command. 

In 7 years of using SSM, I don't think I have ever wanted to run a command without `-x`. The only use case has been for scripts which have made use of `--raw` to build a more complex ssh command. With that in mind, I have modified the `--raw` setting so that it now disables execution.

This PR introduces a new 'dryrun' command line option as a way of getting the old behaviour back (printing out the ssh command rather than executing). 

As this affects the default behaviour of SSM I this will need to be at least a minor version bump and I'll send out some comms as well after releasing.

## What is the value of this?
Two characters less typing every time anyone uses SSM. Only 2 characters! So best not to think about this PR too much or it will take [another year](https://xkcd.com/1205/) to pay off!

## Any additional notes?

### DIFF
--- a/src/main/scala/com/gu/ssm/Main.scala
+++ b/src/main/scala/com/gu/ssm/Main.scala
@@ -11,7 +11,7 @@ object Main {
   implicit val ec: ExecutionContextExecutor = ExecutionContext.global
 
   def main(args: Array[String]): Unit = {
-    val (result, verbose) = argParser.parse(args, Arguments.empty().copy(rawOutput = true)) match {
+    val (result, verbose) = argParser.parse(args, Arguments.empty()) match {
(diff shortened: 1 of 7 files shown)


mchv:
👍 from me

src/main/scala/com/gu/ssm/Main.scala
```diff
@@ -11,7 +11,7 @@ object Main {
   implicit val ec: ExecutionContextExecutor = ExecutionContext.global
 
   def main(args: Array[String]): Unit = {
-    val (result, verbose) = argParser.parse(args, Arguments.empty()) match {
+    val (result, verbose) = argParser.parse(args, Arguments.empty().copy(rawOutput = true)) match {
```
NovemberTang, line 14: Why can't we flip the boolean in the empty() method?

tjsilver (approved):
Great work! 👍

PR #376, opened 2024-01-05
Migrate to scala 3

philmcmahon:
## What does this change?
This PR upgrades ssm-scala onto to scala 3. 

This is my first scala 3 upgrade so I'm by no means an expert, I simply ran the 4 commands outline in the migration plugin https://github.com/scalacenter/scala3-migrate and there were very few changes.

I've tested this locally using both sbt console and the generated executable from `./generate-executable.sh`

## What is the value of this?
I'm not sure - maybe some performance improvements.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-scalaVersion := "2.13.12"
+scalaVersion := "3.3.1"
 
 name := "ssm-scala"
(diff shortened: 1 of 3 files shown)


NovemberTang (approved):
lgtm!

PR #378, opened 2024-01-12
Add `--help` option

davidfurey:
## What does this change?

scopt has a special option with a predefined action called help("help"). When help("help") is defined, parser will print out a short error message when it fails instead of printing the entire usage text.

This is useful, because the existing behaviour is to display the short error message and then the long usage text, which masks the helpful error message.

## What is the value of this?

Less scrolling required to see the helpful error message when you make a mistake running `ssm`

### DIFF
--- a/src/main/scala/com/gu/ssm/ArgumentParser.scala
+++ b/src/main/scala/com/gu/ssm/ArgumentParser.scala
@@ -11,6 +11,8 @@ object ArgumentParser {
 
   val argParser: OptionParser[Arguments] = new OptionParser[Arguments]("ssm") {
 
+    help("help").text("prints this usage text")
+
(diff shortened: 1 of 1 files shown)


philmcmahon (approved):

philmcmahon:
@davidfurey I'm still trying to sort out my last release https://github.com/guardian/homebrew-devtools/pull/97 - happy to do the release for this change and release them together if it would be helpful.

davidfurey:
Sounds good to me, thanks!

Did you mean to close this PR?

philmcmahon:
@davidfurey oh no! No I didn't, sorry

NovemberTang (approved):

PR #406, opened 2024-04-17
Feat: target a specific java version

tjsilver:
## What does this change?

Adds the scala option to target a specific version of the jvm (v11).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,5 +28,9 @@ assemblyMergeStrategy := {
 }
 assemblyJarName := "ssm.jar"
 
-scalacOptions := Seq("-unchecked", "-deprecation")
+scalacOptions := Seq(
+  "-unchecked",
+  "-deprecation",
+  "-release:11",
+)
 


build.sbt
```diff
@@ -28,5 +28,9 @@ assemblyMergeStrategy := {
 }
 assemblyJarName := "ssm.jar"
 
-scalacOptions := Seq("-unchecked", "-deprecation")
+scalacOptions := Seq(
+  "-unchecked",
+  "-deprecation",
+  "-release:11",
```
NovemberTang, line 34: What's the benefit of using this flag?

It's been a while since I've messed about with the JVM, how does this flag interact with us compiling agains Java 21?

tjsilver: IIUC this is something which should have been there all along. Currently when a user runs this program it will run on whatever version of the jvm they have on their machine (and that could theoretically mean it doesn't run). This tells the compiler to make sure that this will work on java version 11 and above. @adamnfish can clarify!

tjsilver: I think it's a separate process from the java 21 on the action, which IIUC just uses a v21 runner to compile and test the app.

adamnfish: All true, but a couple of clarifications!

This is especially relevant because we aren't just making a jar file to publish here, but are actually producing an executable program. This whole thing pointed out a gap in my knowledge, which is how all the below interacts with jars that get published to maven central! I'll have to do some research on it. For now, know that this is all focused on the current use case of creating and executing a java program.


Firstly terminology:
JRE = Java Runtime Environment
  this is the VM that runs jar files
JDK = Java Development Kit
  the developer tools like the compiler (and also includes the JRE)

Both the JDK and JRE are part of released Java versions, 8, 11, 17 and 21 are LTS versions, so they're the main ones we're interested in. 21 is the latest and 8 is now out of date.

Scala compiles your program to a jar file, using the Scala compiler and a [compatible](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html) JDK. Java is releasing new versions regularly these days, but they're still working really hard to make sure everything is backwards compatible. The idea is that newer JDKs can run older jar files, so they don't need recompiling. This means if you compiled something using JDK 8 back in the 2010s it will run flawlessly on a modern JRE (the Java runtime), e.g. 21.

However, these new Java versions introduce new features! This often means a newer JDK can produce more efficient programs, because they can take advantage of new features in the JRE. In particular, the start time for programs on JDK 21 is a lot quicker than it was in previous versions (which makes this the best choice for Lambda functions). But these new features didn't exist in older Java versions! Compiling a program on JDK 21 means it's unlikely to work on JRE 8. The bytecode will be taking advantage of all the new features introduced after version 8, so the v8 JRE won't understand it. In practice, you'll get ClassNotFound or MethodNotFound exceptions at runtime when your JDK 21 program tries to use part of the standard library that don't exist in JRE 8.

To combat this, it's possible to specify the JRE version to target, using the `release` flag. This tells the compiler to produce a jar that only uses features that existed in the specified version. So if you write `release:11`, you are saying only use JRE 11 features. The strong backwards-compatibility guarantees ensure that this jar will work on any later version so it'll be fine on 17, 21 etc.

This is particularly important when we're building an executable that people will run. The release step for this application is to locally build the program and then upload it to Github. From there it gets added to homebrew. The JRE features that get used will depend on whatever version of the JDK is on the engineer that does this laptop. If they have v21, it'll use the v21 features. In this example the engineer would then upload this program to github, and put it in homebrew. But then if another engineer downloads this program onto their laptop that has Java installed at version 17, this is unlikely to work!

The solution is of course, to specify the target version  of Java, using the `release` flag. If we specify `release:11` then compiling the program with Javas 11, 17, 21 or anything else is fine. It'll only use Java 11 features, and therefore will work at version 11 or any higher number. At this point on our timeline, setting `release:11` is ideal!

NovemberTang (approved):

PR #462, opened 2025-01-24
Remove REPL mode from ssm-scala

adamnfish:
## What does this change?

Removes REPL mode support.

## What is the value of this?

This mode of execution was added very early in ssm-scala's life and has never been properly supported or maintained. A cursory experiment suggests this feature is currently broken. The feature brings in additional dependencies, and adds weight too our codebase, so we should remove it if it isn't being used.

The only feature REPL mode offers over and above the standard ssh mode is the ability to run commands on multiple instances at once, but in the modern world I wouldn't expect us to be taking this approach to instance management.


kelvin-chappell (approved):
👍

PR #464, opened 2025-01-24
Scalafmt and compiler warning fixes

adamnfish:
**NOTE:** this PR is currently based on [the remove-repl-mode PR](https://github.com/guardian/ssm-scala/pull/462), since there's no need to fix issues with code that is being removed.

## What does this change?

This is just about tidying things up ahead of future changes.

Introduces scalafmt into the codebase, to stop us having to think about how things are formatted. We also address the compiler warnings in the project.

## What is the value of this?

Tidies things up ahead of updating the AWS Java SDK to v2.

## Any additional notes?

The PR simplifies the implementation of the instance resolution logic, to address an incomplete match warning. Other compiler warnings are mostly about Scala 2.x -> 3.x changes, and have been fixed.

A few deprecation warnings remain relating to the AWS v1 SDK (specifically, the use of `Regions`). The v1 SDK will shortly be replaced with the supported v2 Java SDK, so we can leave these for now.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,2 @@
+version = 3.8.5
+runner.dialect = scala3
--- a/generate-executable.sh
+++ b/generate-executable.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-SCALA_FOLDER="scala-3.6.2"
+SCALA_FOLDER="scala-3.6.3"
 cd $DIR
 sbt assembly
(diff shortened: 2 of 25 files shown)


kelvin-chappell (commented):
Maybe add commit dd5895be06fd91bcf048f2837aba998bf4b3e085 to a [git-blame-ignore-revs](https://gist.github.com/kateinoigakukun/b0bc920e587851bfffa98b9e279175f2) file?

adamnfish:
> Maybe add commit [dd5895b](https://github.com/guardian/ssm-scala/commit/dd5895be06fd91bcf048f2837aba998bf4b3e085) to a [git-blame-ignore-revs](https://gist.github.com/kateinoigakukun/b0bc920e587851bfffa98b9e279175f2) file?

I wasn't aware of this feature, thank you! I will do that when this PR next makes it to the top of my list. For now I'll move this to draft because I have other priorities today and won't get to this until the next 10% day at the earliest.

PR #483, opened 2025-07-07
Upgrade to AWS SDK for Java v2

andrew-nowak:
## What does this change?

Upgrades to the v2 SDK, resolving the warnings on startup that the sdk is approaching eol

## Any additional notes?

PR #492, opened 2025-08-27
retain the services directory when assembling

andrew-nowak:
## What does this change?

Followup to #483 to adjust the assembly merge strategy to handle the AWS SDK v2, which requires us to retain some files in META-INF/services to enable the HTTP clients. As a side bonus, doing so also resolves the SLF4J warnings that "No SLF4J providers were found."

## What is the value of this?

The executable can now run without a stack trace complaining about missing SDK.

## Any additional notes?

We could use MergeStrategy.first instead, like the final fallback case, but deduplicate makes more sense to me, since if the overlapping files aren't identical then we should probably do some investigation to check which one should be kept, rather than assuming the first will be fine. See https://github.com/sbt/sbt-assembly?tab=readme-ov-file#merge-strategy for other options

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,11 +21,14 @@ libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "3.2.19" % Test
 )
 
-// Required as jackson causes a merge issue with sbt-assembly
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
👍

PR #578, opened 2026-07-30
Deprecation

aug24:
## What does this change?

This is the official deprecation of the SSM Scala tool.

It includes documentation on the rationale and worked examples of how the same functionality can be achieved by standard AWS tooling.

ssm-scala-tombstone

## Any additional notes?


deprecation/README.md
```diff
@@ -0,0 +1,119 @@
+# Purpose
+
+AWS SSM has been sufficient to achieve virtually all ssh type needs for some time.  However, it has two
+issues we should address.
+
+## AWS CLI is more powerful
```
adamnfish-gu: Can we put the context lower down? It's interesting to us, but people coming here are going to be focused on how this affects them, which is only "there's an AWS command that does the same thing".

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+port forwarding, and other useful ssh features.
+
+The AWS CLI has since been updated to include almost the same functionality, but better.  Instances no longer need to
+be permitted access via the ssh port, and the AWS CLI can now do port forwarding by host name.
+
+SSM Scala is no longer worth the maintenance cost.  It won't be deleted, but you should consider moving away from it.
+
+![ssm-scala-tombstone.png](ssm-scala-tombstone.png)
+
+## More paranoia around credentials
```
adamnfish-gu: This section is also not relevant to people's day-to-day use. Do we really need to say any of this here?

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+# Fetch the private cert
+aws --region eu-west-1 s3 cp s3://my-bucket-of-secrets/my-stack/PROD/my-secure-app/secret-cert.json
+```
+or EVEN
+```
+curl -H "X-aws-ec2-metadata-token: $TOKEN" \
+  http://169.254.169.254/latest/meta-data/iam/security-credentials/[role-name]
+```
+
+Incidentally, there is now a dedicated `.ssh` permission in Janus, which gives just enough permissions to discover
```
adamnfish-gu: This starts with "incidentally", but is it actually the first non-incidental part of the document?

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Even with those reduced credentials, we should aim use the same approach as devcontainers, and not put "bare" 
+credentials on our laptops (where they can potentially be exfiltrated by attackers) at all.
+
+This is a simple setup to deliver containerised ssh to instances, with discover built in.  You can use this, or take
+the principles and build your own, but the important thing is to avoid putting credentials on your laptop.
+
+# Setup
+
+Build a small docker image which includes some scripts, the AWS cli tool and, crucially, the SSM session manager plugin for AWS.
```
adamnfish-gu, line 7: Is this indirection actually necessary? Could we just give people the AWS CLI command, and document any requirements?

Having done that, and only if really necessary, we could give this tool.

I think we should also be saying "use the AWS console", to avoid almost all this complexity. The AWS Console has a terminal in it, and means we can eliminate all this complexity. We don't need to set up dependencies on the machine, we don't need to resolve the instance from tags. We say to people "go to the console, find the instance, click connect".

As we heard in the presentation, I don't think people really want a separate docker container with these scripts in!

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+```
+docker build -t aws-shell .
+```
+
+# Run
```
adamnfish-gu: As above, I think this is a lot of indirection for achieving our goal, can we try and simplify all of this?

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+Grab your AWS credentials as `aws configure` commands.  Make sure the profile name is `default` or set `AWS_PROFILE`.
+Paste them into the docker shell.  The region is already set to eu-west-1, but can be overridden on the command line.
+
+# Examples
+
+## Starting a remote SSM session
+
+```
+docker run -it aws-shell bash
```
adamnfish-gu, line 31: The example ties users to the docker container you used for carrying an isolated example. As above, I don't think people want to work this way in practice. Could we distill the docs to their essence (or avoid needing this at all, e.g. by recommending the console)

deprecation/README.md
```diff
@@ -0,0 +1,119 @@
...
+Connecting to i-0c7f9cb1234567890
+
+Starting session with SessionId: justin.rowles-rycskjdsfjkgflkbgdflksdfakl
+```
+
+## Starting a tunnel to a remote host
+
+### Ports
+
+There will be three ports in play.
```
adamnfish-gu, line 46: Again, we're meaningfully contributing to the complexity here, from our design. Perhaps we can prioritise simplicity?

I think this section is likely the only bit that where we need to be documenting the local process, since this is inherently a "local machine" concern.

But on the assumption people should be using devcontainers anyway going forwards, is there a more direct approach?

README.md
```diff
@@ -1,3 +1,27 @@
+Deprecation Announcement
+========================
+
+SSM-Scala is now considered deprecated.  All its functionality is available within the AWS CLI.  eg:
```
adamnfish-gu: Can we also recommend the AWS Console's "connect" button, which I think is going to have a lot less operational complexity. This also enforces better isolation, since people then wouldn't be running it on their laptops.

deprecation/ssm-scala-tombstone.png
kelvin-chappell, line 1: That's a big picture!  But I don't suppose many people are going to download it.

README.md
```diff
@@ -1,3 +1,27 @@
...
+    --filters \
+        "Name=tag:App,Values=$app" \
+        "Name=tag:Stack,Values=$stack" \
+        "Name=tag:Stage,Values=$stage" \
+        "Name=instance-state-name,Values=running" \
+    --query "Reservations[].Instances[] | sort_by(@, &LaunchTime)[0] | InstanceId"
+```
+and starting a session on an instance:
+```
+aws ssm start-session --target $instanceid
```
adamnfish-gu: let's add the profile to both these examples, since we overwhelmingly use AWS profiles to manage access.

e.g.
```bash
aws ssm start-session --target $instanceid --profile <profile>
```

README.md
kelvin-chappell, line 1: The section https://github.com/guardian/ssm-scala/blob/507f633f431a0223d8c160a53292ce8ff161397d/README.md#modern-alternatives seems to be redundant now that you've provided detailed alternatives.

README.md
```diff
@@ -1,3 +1,27 @@
...
+```
+aws ec2 describe-instances \
+    --filters \
+        "Name=tag:App,Values=$app" \
+        "Name=tag:Stack,Values=$stack" \
+        "Name=tag:Stage,Values=$stage" \
+        "Name=instance-state-name,Values=running" \
+    --query "Reservations[].Instances[] | sort_by(@, &LaunchTime)[0] | InstanceId"
+```
+and starting a session on an instance:
```
kelvin-chappell: Perhaps it's worth saying that we recommend finding alternative ways to do things other than going onto boxes.  That this should be considered a last resort.   And I suppose if you have to do it it's best to work in the AWS console as far as possible.

adamnfish-gu (commented):
I think this looks quite complicated! We're deferring to an AWS feature, so it'd be nice to lean more heavily on them.

Firstly, I think we should be recommending the AWS Console's Connect button because it is safer and simpler. Much of the complexity we have here goes away if we encourage people to use the AWS Console to handle the administrative edge case of needing to get a shell on an instance.

We can't recommend that for a local tunnel, so we'll need some documentation for using the CLI. I think bringing this like an entire tool with its own docker container etc is only moving the problem of maintaining ssm-scala rather than solving it. We expect people to be working in containers anyway, and where their workflow depends on SSH tunnels, it'll be up to them to make sure their dev environments meet those requirements. If there are things to install they can do that there and we could even offer a devenv module to help - if this is widely needed. It's an if because we really do want to try to discourage this approach in general.

I'd also observe that this reads like it was written by and for us (DevX/Security), rather than the target audience. Can we prioritise the information people that visit this in the future will want to read, which is how they can get a shell to their instance. The background is nice, and some people will be interested, but it shouldn't be the primary focus, as now.

kelvin-chappell (approved):
👻 
Just a couple of comments for the readme.

PR #589, opened 2026-08-31
add dependabot grouping, and sbt dependency management

NovemberTang:
## What does this change?

Groups nonbreaking dependabot updates to save time
Adds SBT as a dependabot option (available since may). Let's see how this behaves and then consider removing scala steward


kelvin-chappell (approved):
👍 but isn't this repo about to be archived?

